PR #45, opened 2023-01-29
Add Japanese localization

soya-daizu:
Hi, I just found and tried this app, and it's working really great! I was waiting for an open-source, cross-platform alternative to AirDrop like this so that I can move files between my Linux laptop and iOS devices.

I made a Japanese localization for the app and I hope it gets merged. I don't have Flutter installed so I haven't tested it in real app, but I was able to look around and find where each string is used, so I believe the translations are not out of context.


Tienisto:
Thanks!

PR #58, opened 2023-01-29
feat(i18n): add fr_FR translation

florian-lefebvre:
I may have forgotten to update a file, let me know and I will update the PR.


Tienisto:
Thank you! You don't need to update `_missing_translations.json` btw. It will be generated by a tool.

PR #61, opened 2023-01-29
feat(i18n): add it translation

czar0:
Added `it` (Italian) translation


Tienisto:
Thanks! You don't need to update `_missing_translations.json`. It will be updated automatically

PR #65, opened 2023-01-30
added `star-history`

hemangjoshi37a:


Tienisto:
Man I don't care about stars. Just use this app if you like, if not, don't

PR #76, opened 2023-01-31
Added sv translation

davidbergqvist:
Added Swedish translations. Was not sure if _missing_translations.json had to be updated, so I did just in case.


Tienisto:
Thank you! The `_missing_translations.json` file is not needed.

PR #83, opened 2023-01-31
fix: multiline property of message input dialog

### LINKED ISSUE #73: Multiline checkbox for text input not working
Inside the "type message" dialog checking or unchecking the "multiline" checkbox does not change the behaviour of the field.
Either way, when typing a long string, the dialog is expanded to the max width of the screen (with some padding) and after that a multiline is created.

I am testing on Android 11 on an OnePlus 6.

S-ecki:
### Goals
- multiline checkbox of message input dialog changes behavior of text field, making it horizontally scrollable when disabled

closes #73


Tienisto (approved):

Tienisto:
Nice thank you!

PR #97, opened 2023-02-01
Add strings file for spanish (Spain).

alemalvarez:
This pull request adds strings_es-ES.i18n.json file with a translation to Spanish. I think they are mostly correct. However I wasn't able to run the app (Flutter refuses to work for me), so I haven't seen them in context. 
It needs to be noted that the alias generator won't make much sense in Spanish as adjectives change depending on the noun but it's better than nothing.

### DIFF
--- a/assets/i18n/strings_es-ES.i18n.json
+++ b/assets/i18n/strings_es-ES.i18n.json
@@ -50,14 +50,14 @@
     "selection": {
       "title": "Selección",
       "files": "Ficheros: {files}",
-      "size": "Tamaño:: {size}"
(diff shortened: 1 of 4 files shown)


Tienisto:
Hi, you can leave the alias generator out. It will use the English version. You can look at the German (de) translations for example.

alemalvarez:
Now it is removed. I also fixed some small typos.

Tienisto:
You should rebase your branch onto the latest version of  `main`. If this is too difficult, then you can also close this PR and create a new one (and copy the spanish translations).

PR #44, opened 2023-01-28
Create strings_zh-Hant_TW.i18n.json

virgil724:
Use https://zhconvert.org/ to translate from zh-Hans and include mine modifications

PR #103, opened 2023-02-01
Improve Traditional Chinese Translation

graphemecluster:
#44 used https://zhconvert.org/ to convert individual characters from simplified to traditional, but none or few differences in vocabulary and terms are taken into consideration. https://opencc.byvoid.com/ did a great job on the conversion.


Tienisto (approved):
Thanks!

PR #104, opened 2023-02-01
Traditional Chinese (Hong Kong) Translation

graphemecluster:
This translation is based on the previous PR, while taking both language and vocabulary differences into consideration.


Tienisto (approved):
Thank you!

PR #110, opened 2023-02-01
Update Japanese locale

soya-daizu:
We should probably discuss about it in another place, but have you considered using a translation management service like Crowdin, Weblate, POEditor, Transifex, and etc? While they are mostly paid services, some of them also offer free tiers for open source projects.

### DIFF
--- a/assets/i18n/strings_ja.i18n.json
+++ b/assets/i18n/strings_ja.i18n.json
@@ -93,7 +93,9 @@
       "needRestart": "設定を反映するにはサーバーを再起動してください！",
       "server": "サーバー",
       "alias": "エイリアス",
-      "port": "ポート"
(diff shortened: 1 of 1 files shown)


Tienisto (approved):

Tienisto:
Thanks for your contribution!

Currently, there is no need for an external service because the i18n library already provides good analysis of missing and unused translations.

These are stored in (assets/i18n/_missing_translations.json)[https://github.com/localsend/localsend/blob/main/assets/i18n/_missing_translations.json] and (assets/i18n/_unused_translations)[https://github.com/localsend/localsend/blob/main/assets/i18n/_unused_translations.json].

soya-daizu:
Alright, I understand your approach. Thanks for explaining it!

PR #118, opened 2023-02-03
Update strings_it.i18n.json

albanobattistella:


Tienisto:
Note: This is not needed because entries starting with `@` are comments

PR #122, opened 2023-02-04
feat: add pt-BR translations

KyleKun:
Hi @Tienisto, first of all, thanks so much for this app 😊

I've added Brazilian Portuguese strings, please let me know if any adjusts are needed.


Tienisto:
Nice, thank you!

PR #182, opened 2023-02-19
Update strings_hu.i18n.json

gidano:


assets/i18n/strings_hu.i18n.json
```diff
@@ -169,7 +169,7 @@
     "title": "Változásnapló"
   },
   "aliasGenerator(ignoreMissing)": {
-    "@info": "Inherits from the English version"
+    "@info": "Az angol változatból örökölte"
```
Tienisto, line 172: Comments should not be translated

gidano: Ok. Did you fix it or should I do it and add another pull request?

Tienisto: This is not a major issue, so I just merge it :)

PR #230, opened 2023-03-03
Impoving russian localization

Pr0gger1:
Russian localization was improved. Fixed words mistakes and punctuation


Tienisto:
Thank you!
If you like, you can also add missing translations: https://github.com/localsend/localsend/blob/main/assets/i18n/_missing_translations_ru.json

PR #255, opened 2023-03-07
i18n: update to pt-br translation

TheGB0077:
Cleared the latest missing translations backlog and applied translations with slang apply to strings_pt-br. 

Changes to take be aware:

- The recentlyUsed field has been translated to "Recém conectado" ( Recently connected ), that implies a list of recently IP addresses that had a successful connection.

Note: The "(Download)" string translates 1:1 to Portuguese, so it was applied as it is.


Tienisto:
Thank you!

PR #286, opened 2023-03-13
i18n update & File not found dialogue tweak

TheGB0077:
Small changes to bring a more natural feel to the reading experience.

### DIFF
--- a/lib/widget/dialogs/cannot_open_file_dialog.dart
+++ b/lib/widget/dialogs/cannot_open_file_dialog.dart
@@ -3,13 +3,14 @@ import 'package:localsend_app/gen/strings.g.dart';
 import 'package:localsend_app/util/platform_check.dart';
 import 'package:localsend_app/widget/dialogs/custom_bottom_sheet.dart';
 import 'package:routerino/routerino.dart';
+import '../../provider/receive_history_provider.dart';
 
 class CannotOpenFileDialog extends StatelessWidget {
   final String path;
 
   const CannotOpenFileDialog({required this.path, super.key});
 
-  static void open(BuildContext context, String path) {
+  static void open(BuildContext context, String path, String? fileId, ReceiveHistoryNotifier? filesRef) {
     if (checkPlatformIsDesktop()) {
       showDialog(
         context: context,
@@ -17,6 +18,10 @@ class CannotOpenFileDialog extends StatelessWidget {
           title: Text(t.dialogs.cannotOpenFile.title),
           content: Text(t.dialogs.cannotOpenFile.content(file: path)),
           actions: [
+            TextButton(
+            onPressed: () => { filesRef!.removeEntry(fileId!), context.pop() },
+            child: Text(t.receiveHistoryPage.entryActions.deleteFromHistory),
+            ),
             TextButton(
               onPressed: () => context.pop(),
               child: Text(t.general.close),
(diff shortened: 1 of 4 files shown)


TheGB0077:
Closes #287

lib/widget/dialogs/cannot_open_file_dialog.dart
```diff
@@ -3,20 +3,25 @@ import 'package:localsend_app/gen/strings.g.dart';
...
+  static void open(BuildContext context, String path, String? fileId, ReceiveHistoryNotifier? filesRef) {
     if (checkPlatformIsDesktop()) {
       showDialog(
         context: context,
         builder: (_) => AlertDialog(
           title: Text(t.dialogs.cannotOpenFile.title),
           content: Text(t.dialogs.cannotOpenFile.content(file: path)),
           actions: [
+            TextButton(
+            onPressed: () => { filesRef!.removeEntry(fileId!), context.pop() },
```
Tienisto, line 22: In Dart/Flutter, `=>` is used for single statements. You are returning a `Set` containing 2 items here.

You need something like this:

```dart
onPressed: () {
  a();
  b();
}
```

lib/widget/dialogs/cannot_open_file_dialog.dart
```diff
@@ -3,20 +3,25 @@ import 'package:localsend_app/gen/strings.g.dart';
...
-  static void open(BuildContext context, String path) {
+  static void open(BuildContext context, String path, String? fileId, ReceiveHistoryNotifier? filesRef) {
     if (checkPlatformIsDesktop()) {
       showDialog(
         context: context,
         builder: (_) => AlertDialog(
           title: Text(t.dialogs.cannotOpenFile.title),
           content: Text(t.dialogs.cannotOpenFile.content(file: path)),
           actions: [
+            TextButton(
```
Tienisto, line 21: Since removing the file from history is optional, you can make this button optional too.

```dart
actions: [
  if (fileId != null)
    TextButton(
       // ...
     ),
]
```

TheGB0077: But would fileId ever be null? The dialogue can only appear if you click a file entry in the history, so there will always be a file entry to be deleted with this action, or am I mistaken?

lib/widget/dialogs/cannot_open_file_dialog.dart
TheGB0077: The Sets approach worked, but I agree that is not the most straightforward. I'll commit the changes!

lib/widget/dialogs/cannot_open_file_dialog.dart
Tienisto: As a sender, you can open the file before sending the file to the recipient (see `selected_files_page.dart`).

lib/widget/dialogs/cannot_open_file_dialog.dart
Tienisto: Yep, during `Set` creation, the necessary method calls are taken. That's why it worked :)

lib/widget/dialogs/cannot_open_file_dialog.dart
TheGB0077: > As a sender, you can open the file before sending the file to the recipient (see `selected_files_page.dart`).

I see, I previously asserted the values as non-null thinking otherwise.

Tienisto (approved):

Tienisto:
Nice thank you!

Issue #296, opened 2023-03-17
remember Window size and position


bachig26:
### Is your feature request related to a problem? Please describe.

From my experience, I like to keep/organize the app position and size relatively along the other apps side-by-side for a multitasking perspective, but every new session relapses over the existing app/tool that I've been using and the position that I'm not comfortable with.

I use a lot with localsend from the point when I found this app, an I want a consistent and less distractive workflow, so I hope to see this feature in play.

### Describe the solution you'd like

This issue is to request an function to remember the window size and position of the localsend in desktop, also an option in the settings to toggle them.

### Additional context

_*this feature is for desktop version._

electr1fy0:
Adding to this, I still believe the Mac app should be a Menu Bar app only.

TheGB0077:
I'll try to sketch up the code for this! @bachig26

Tienisto:
Thanks @TheGB0077!

You will probably need to store some keys in `PersistenceService`. Then you can write/read those keys during init (https://github.com/localsend/localsend/blob/main/lib/init.dart#L91)

TheGB0077:
Just pushed a draft! 
I hardcoded a variable to replace the setting for now, although the UI implementation should be straightforward enough.

bachig26:
@TheGB0077 @Tienisto is it possible include the stick to edge function along this ongoing draft?

bachig26:
> In the next version of the app? I think it's likely. 

❤

The only thing yet to consider is if the settings option is really necessary, at least considering that having the window restore the previous position/size is a great default, so is it really necessary to let people opt out? What do you think @bachig26 ?

on second thought, I'm not sure about an option to toggle it in settings, as from my perspective I want it to be active all the time.
I don't know maybe if you want and can try it out and see if you need the toggle or not, then go for the one you like it.

Tienisto:
I think having a settings option is not necessary. I have never seen this option on another desktop app (maybe because I didn't cared)

sergd88:
On the contrary, I like the fact that the default location for program window is always center, this is very convenient, especially since the application supports drag-and-drop, so you can freely move the window, change its size, knowing that the next time you open app, window will be fixed in size and in the center.

sergd88:
> I think having a settings option is not necessary. I have never seen this option on another desktop app (maybe because I didn't cared)

At first glance, there is a similar function in Total Commander. Although it is somewhat different, you can manually save the window size and position by selecting the appropriate option in the settings menu.

[...]

PR #304, opened 2023-03-20
feat: Restore window size and position

TheGB0077:
Initial setup to enable restoring the position and size of the app! 

- Setting up Persistence for the required App Window data ☑
- Changes to `init.dart` ☑
- Interface changes to accommodate the option to allow the feature ☒
- Setting up Persistence for the boolean config value ☒

Currently the information is saved on the window watcher onWindowClose method but, perhaps, some other method would be more suitable.

Solves #296

### DIFF
--- a/lib/provider/persistence_provider.dart
+++ b/lib/provider/persistence_provider.dart
@@ -1,4 +1,5 @@
 import 'dart:convert';
+import 'dart:ui';
 
 import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
@@ -67,11 +68,6 @@ class PersistenceService {
       await prefs.setString(_themeKey, ThemeMode.system.name);
     }
 
-    return PersistenceService._(prefs);
-  }
-
-  static Future<PersistenceService> getPersistence() async {
-    final prefs = await SharedPreferences.getInstance();
     return PersistenceService._(prefs);
   }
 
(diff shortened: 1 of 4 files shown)


lib/provider/persistence_provider.dart
```diff
@@ -64,6 +70,11 @@ class PersistenceService {
     return PersistenceService._(prefs);
   }
 
+  static Future<PersistenceService> getPersistence() async {
```
Tienisto, line 73: `riverpod` is used as the dependency injection tool so you can get the instance by calling

`final persistenceService = ref.watch(persistenceProvider);`

The widget must be a `ConsumerWidget` / `ConsumerStatefulWidget` though. So either you can change the `WindowWatcher` from a `StatefulWidget` to a `ConsumerStatefulWidget` or you can get the `persistenceService` inside `main.dart`.

Tienisto:
Listening to the windows close event is good enough. Nice!

lib/provider/persistence_provider.dart
TheGB0077: Noted! I'll rewrite a couple of things and make a new commit soon.

Tienisto:
Thinking about the settings again: I don't know if this is overkill. We can just save the last window position/size and that's it.

TheGB0077:
Oh yeah, originally I intended to discuss this.
 I thought that the setting could be implemented as enabled by default and opted out by the few that prefer to have it off, although it is indeed a bit of a reach to think the restoration of the window position/size as a nuisance to anyone. 
Besides, implementing it would imply on creating some extra I18n strings that would've to be translated by everyone and that could be avoided.

Tienisto:
Thanks! It works!

PR #317, opened 2023-03-24
i18n: Update translations for Ukrainian and Russian language (uk-UK; ru-Ru)

sergd88:
Translated missing additional fragments and made some changes and corrections to the existing translation (for example: line 215 @:general.error, which is not translated in other languages and is used in English, has also been reverted back to English).


sergd88:
Update translation for Russian language (ru-Ru)

PR #324, opened 2023-03-25
Asset Picker default Text delegate replaced

TheGB0077:
Enables extension of languages supported by the library.


lib/util/asset_picker_translated_text_delegate.dart
```diff
@@ -0,0 +1,94 @@
+import 'package:localsend_app/gen/strings.g.dart';
+import 'package:wechat_assets_picker/wechat_assets_picker.dart';
+
+
+class TranslatedAssetPickerTextDelegate extends AssetPickerTextDelegate {
+  const TranslatedAssetPickerTextDelegate();
+
+  @override
+  String get languageCode => LocaleSettings.useDeviceLocale().languageCode;
```
Tienisto, line 9: You can use this: `LocaleSettings.currentLocale.languageCode` to avoid side effects from `useDeviceLocale`

TheGB0077:
Added all the missing translations files from the previous pull request, except Ukrainian and Russian to avoid conflicts with #325

lib/util/asset_picker_translated_text_delegate.dart
TheGB0077: Done

TheGB0077:
Oh, and that remaining untranslated string from the German file really wasn't nowhere to be found in the original locales from the lib for some reason.

Tienisto (approved):

Tienisto:
Nice!

PR #325, opened 2023-03-25
i18n: Update translations for Ukrainian and Russian language (uk-UK; ru-Ru)

sergd88:
The translation of the line 89: saveWindowPlacement for [323](https://github.com/localsend/localsend/pull/323), also added a translation for  missing_translations, which for some reason didn't save last time.


Tienisto:
The `@@info` and other `@` strings should not be translated as they are comments and not used in the app. `@@info` in particular, is generated by the tool.

sergd88:
> The `@@info` and other `@` strings should not be translated as they are comments and not used in the app. `@@info` in particular, is generated by the tool.

Well, the @@info line, I didn't translate. At the initial stage, when I first translated, I translated the line @:general.error, but then I realized that it was some kind of markup for comments and returned it as it was.

PR #329, opened 2023-03-26
Save dimensions on resize and move on linux

TheGB0077:
Solved the last edge case I could think about + added some handy comments :)

### DIFF
--- a/lib/provider/window_dimensions_provider.dart
+++ b/lib/provider/window_dimensions_provider.dart
@@ -1,29 +1,37 @@
 import 'dart:ui';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:localsend_app/provider/persistence_provider.dart';
 import 'package:screen_retriever/screen_retriever.dart';
 import 'package:window_manager/window_manager.dart';
 
 //Records are a better alternative, but they are currently experimental
-typedef WindowDimensions = Map<String, OffsetBase?>?;
+typedef WindowDimensions = Map<String, OffsetBase?>;
 
-class WindowDimensionProvider {
+final windowDimensionProvider = Provider<WindowDimensionsController>((ref) {
+  return WindowDimensionsController(ref.watch(persistenceProvider));
+});
+
+class WindowDimensionsController {
   final PersistenceService _service;
   static const Size minimalSize = Size(400, 500);
   static const Size defaultSize = Size(900, 600);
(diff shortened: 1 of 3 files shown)


lib/widget/watcher/window_watcher.dart
```diff
@@ -52,23 +53,39 @@ class _WindowWatcherState extends ConsumerState<WindowWatcher> with WindowListen
...
 
+  //Linux alternative for onWindowMoved and onWindowResized
+  @override
+  Future<void> onWindowMove() async {
+    if(checkPlatform([TargetPlatform.linux]) && !timeout) {
+      timeout = true;
+      final windowOffset = await windowManager.getPosition();
+      final windowSize = await windowManager.getSize();
+      await _dimensionsController?.storeDimensions(windowOffset: windowOffset, windowSize: windowSize);
+      Timer(const Duration(milliseconds: 600), () => toggleTimeout());
```
Tienisto, line 64: I think it is better to store the end timestamp instead of a boolean so you don't need to create a timer.

lib/provider/window_dimensions_provider.dart
```diff
@@ -1,29 +1,37 @@
...
 import 'package:localsend_app/provider/persistence_provider.dart';
 import 'package:screen_retriever/screen_retriever.dart';
 import 'package:window_manager/window_manager.dart';
 
 //Records are a better alternative, but they are currently experimental
-typedef WindowDimensions = Map<String, OffsetBase?>?;
+typedef WindowDimensions = Map<String, OffsetBase?>;
 
-class WindowDimensionProvider {
+final windowDimensionProvider = Provider<WindowDimensionsController>((ref) {
```
Tienisto, line 10: Yep, this is the riverpod-way. Nice :)

lib/widget/watcher/window_watcher.dart
TheGB0077: Used a Stopwatch for overall better syntax with elapsedtime.

lib/provider/window_dimensions_provider.dart
TheGB0077: Incrementally we got there! I was testing things out with the StateController concept with currentDimensions, but ultimately I opted to just write up the Provider proper and be done with it 😂

Tienisto (approved):

PR #351, opened 2023-04-01
fix: Alias generation properly localised

TheGB0077:
I noticed the Alias generation function was called before the actual locale info being set, thus the app was never aware the of the language it should generate the Alias, but with this fix it works fine!

image


Tienisto (approved):

Tienisto:
Nice catch!

PR #360, opened 2023-04-05
avoid creating text files when sending simple text links or messages

jonatino:
Transfer text as text instead of text files. Perhaps we can decide on a character count at which point text will be sent as a .txt file (perhaps > 4000 characters similar to discord), otherwise send the message as plain text.

Will open later PR's for refactoring the way the history/receive page looks. Also will add support for one click opening links if a url is detected in plain text message


Tienisto:
I think there is no need to add a new type `utf8Raw`. We can just leave the "preview" out so the recipient has to download the file.

PR #368, opened 2023-04-07
Ios Local Network Permission

TheGB0077:
This is a potential fix for the Local Network Permission not showing up. 

The approach is the same as found in here: [doozMen Code](https://gist.github.com/doozMen/0b5fc54c765bccb7c13792caa4eaa51c)  

If need may be, the [Capacitor Code](https://github.com/luminsmart/capacitor-permissions-checker/blob/1b4a1609758941709b930b902f0c38fdadefbab6/ios/Plugin/LocalNetworkAuthorization.swift) could be used instead.

#### The changes need to be tested still (as I do not have a Mac / iPhone), thus I cannot verify for myself if the changes had the desired effect. 

If it does not work, I'll revert the changes, except for the addition of intent to use the Local Network on the **plist file**, such that is required by Apple.

#### ` Future changes `: I took notice that dio has a [NativeAdapter](https://github.com/cfug/dio/tree/main/plugins/native_dio_adapter) in the works, but it looks like it's still not ready for production, so it could be something to consider in the future to make use of Native APIs from macOS, iPhone and Android.

### DIFF
--- a/assets/i18n/strings.i18n.json
+++ b/assets/i18n/strings.i18n.json
@@ -341,6 +341,11 @@
       "single": "Sends files to one recipient. Selection will be cleared after finished file transfer.",
       "multiple": "Sends files to multiple recipients. Selection will not be cleared.",
       "link": "Recipients who do not have LocalSend installed can download the selected files by opening the link in their browser."
+    },
+    "localNetworkUnauthorized": {
+      "title": "Local Network discovery unauthorized",
+      "description": "Localsend can't find other devices without this permission, so make sure it's enabled!",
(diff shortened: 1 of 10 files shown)


TheGB0077:
Testing out the dialog right now!

Screenshot_20230415-230724

Tienisto:
I just tested this on a real device (iOS 16) and indeed, your changes fixes this problem. Thanks alot!

LGTM. You can remove the draft status if you are done.

Tienisto (approved):

Tienisto:
I will just merge it! You can create another PR if you have some improvements!

PR #388, opened 2023-04-17
Web UI styles added and small fixes

TheGB0077:
#### So far only the Teal color has been used from Localsend's color palette, but the page could incorporate more from it. Also, the styles could be split to a separate file, if necessary.

Other Changes: 

```javascript
	// Looking for "Win" can match with Windows, Win64 or Win32. 
	if (userAgent.contains("Windows")) {
    if (userAgent.contains("Win")) {
	
	// iOS userAgent also contains Mac OS, example: 
	// Mozilla/5.0 (iPhone13,2; U; CPU iPhone OS 14_0 like Mac OS X) AppleWebKit/602.1.50 (KHTML, like Gecko) Version/10.0 Mobile/15E148 Safari/602.1
    } else if (userAgent.contains("Macintosh") || userAgent.contains("Mac OS")) {
    } else if (userAgent.contains("Macintosh")) {

	// Using "X11" ensures it is a Linux desktop and not a potential variant, like Android.
    } else if (userAgent.contains("Linux")) {
    } else if (userAgent.contains("X11"))

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ You can help translating this app to other languages!
 ### Bug fixes and improvements
 
 - **Bug Fixes:** If you find a bug, please create a pull request with a clear description of the issue and how to fix it.
-- **Improvements:** Have an idea for how to improve LocalSend? Please create an issue first so we can discuss why the improvement is needed.
+- **Improvements:** Have an idea for how to improve LocalSend? Please create an issue first, so we can discuss why the improvement is needed.
 
 For more information, see the [contributing guide](https://github.com/localsend/localsend/blob/main/CONTRIBUTING.md).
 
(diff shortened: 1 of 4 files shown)


Tienisto:
The text over the teal background seem to be hard to read. Maybe use a lighter teal color?

TheGB0077:
The 2-column separation only shows up on a tablet size view, because if there is only 1, then the row size gets stretched way too much.
On a regular phone there's only 1 column

image

image

And then at the desktop level we have a regular view
image

#### And what about this tone, is it good?

image

Tienisto:
I see. Having the file size as 2nd row still seems to be overkill as there is only one attribute (the size).
Maybe have 4 columns instead so we have:

|File name|size|file name|size
|---|---|---|---|
|file a|123|file b |123

Regarding the color, you can take the 300-500 version of `teal`: https://materialui.co/colors/

Tienisto:
Another solution for tablet is to write the file size in brackets: `myImage.jpeg (3.2 MB)`. So you don't need a second row either.

Tienisto:
You could also add the teal color to the table header, then you don't need to have it in the file size column

TheGB0077:
> Another solution for tablet is to write the file size in brackets: `myImage.jpeg (3.2 MB)`. So you don't need a second row either.

I did with a::after 
image

image


image

TheGB0077:
This is how it looks with Teal 400

image

Tienisto:
Yeah, I am fine with that 👍

Tienisto:
Nice, thanks!

PR #392, opened 2023-04-17
fa (farsi-persian) Translation

farshad991:


assets/i18n/strings_fa.i18n.json
```diff
@@ -0,0 +1,400 @@
...
+      "هلو",
+      "گلابی",
+      "آناناس",
+      "سیب زمینی",
+      "کدو تنبل",
+      "تمشک",
+      "توت فرنگی",
+      "گوجه"
+    ],
+    "combination": "{adjective} {fruit}",
```
Tienisto, line 277: Just to be sure: The combination `adjective - fruit` is correct and the adjectives and fruits can be combined without any modifications? Some languages may require modifications to the adjective depending on the fruit.

PR #404, opened 2023-04-20
feat: linux tray enabled

TheGB0077:
This change needs some careful considerations...

Unfortunately to incorporate this into the codebase the Linux workflow changes to some degree:

- First of all Flutter has to be installed manually, the snap version has packages included that are not compatible, so we'll need to replace it:  (see https://github.com/leanflutter/tray_manager/issues/28#issuecomment-1308688667)
- The PATH variable will need to be set for flutter and is most reliably set by changing ~/.bashrc manually
- Extra dependencies: lib32stdc++-12-dev, libayatana-appindicator3-dev and the package [system_tray](https://pub.dev/packages/system_tray) 
(I tried to make the tray manager work for Linux but it didn't work still)
- flutter clean will likely be necessary before trying to run the app again

Some extra troubleshooting might be necessary, but those are the steps I took.

### DIFF
--- a/lib/util/native/platform_check.dart
+++ b/lib/util/native/platform_check.dart
@@ -17,7 +17,6 @@ bool checkPlatformIsDesktop() {
 }
 
 /// This platform supports tray
-/// On linux, this is currently not supported
 bool checkPlatformHasTray() {
   return checkPlatform([TargetPlatform.windows, TargetPlatform.macOS, TargetPlatform.linux]);
 }
--- a/lib/util/native/platform_check.dart
+++ b/lib/util/native/platform_check.dart
@@ -19,7 +19,7 @@ bool checkPlatformIsDesktop() {
 /// This platform supports tray
 /// On linux, this is currently not supported
 bool checkPlatformHasTray() {
-  return checkPlatform([TargetPlatform.windows, TargetPlatform.macOS]);
+  return checkPlatform([TargetPlatform.windows, TargetPlatform.macOS, TargetPlatform.linux]);
 }
 
 /// This platform can receive share intents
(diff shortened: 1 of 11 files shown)


lib/util/native/tray_helper.dart
```diff
@@ -14,31 +15,59 @@ Future<void> initTray() async {
...
+      iconPath: Assets.img.logo32Png.path,
+    );
+
+    final Menu menu = Menu();
+    await menu.buildFrom([
+      MenuItemLabel(label: t.tray.open, onClicked: (menuItem) async {
+        await windowManager.show();
+        await windowManager.focus();
+    }),
+      MenuItemLabel(label: t.tray.close, onClicked: (menuItem) => windowManager.destroy()),
```
Tienisto, line 59: You can use `exit(0)` for that so we don't rely on a library.

lib/util/native/platform_check.dart
```diff
@@ -19,7 +19,7 @@ bool checkPlatformIsDesktop() {
 /// This platform supports tray
 /// On linux, this is currently not supported
```
Tienisto, line 20: I guess this comment is obsolete now

TheGB0077: The check itself has become redundant as it is the same as `checkPlatformIsDesktop`, but I guess that it gives some extra context to the code that it belongs to, so I don't think it needs to be removed.

TheGB0077:
> Thanks. I expected a more ugly change with more C code but this PR seems ok!

Earlier I did say the solution wasn't pretty, but I was more concerned about the snap ordeal. 
Installing flutter manually shouldn't be a road blocker for most Linux people, but the snap installation looks like it is the preferred one by the flutter ecosystem.

lib/util/native/tray_helper.dart
TheGB0077: Done

Zeno394:
LocalSend v 1.8 runs well on MX Linux but latest version doesnt work.
Thanks

PR #475, opened 2023-05-19
Fix/receiver unable to cancel send session

SelaseKay:


Tienisto:
Looking good overall. Nice!

I will add some security checks, so not anyone can cancel the session.

SelaseKay:
Awesome!!!

PR #483, opened 2023-05-25
Fix: Incorrect position set on initialization

TheGB0077:
Method isInScreenBounds implemented to detect if the last known position is valid.


Tienisto (approved):

Tienisto:
Nice!

PR #500, opened 2023-06-03
i18n: Update translations for Ukrainian, Russian and English language

sergd88:


sergd88:
The term "Brightness" is a somewhat complex formulation that essentially describes the "Theme" of the system, as named by the OS developers.

Tienisto:
Thank you!

PR #542, opened 2023-06-22
feat: use Escape key to quit about and changelog page

RiverTwilight:
This change only affects two pages: about and changelog.

After use press Escape key on their keyboard will back to previous page. 

I add a `ShortcutWrapper` for the `Scaffold` component of these two pages to achieve this.


Tienisto:
We can actually make this global. I will merge this and make some adjustments.

Issue #529, opened 2023-06-11
Start minimized not working in Windows


Tienisto:
It is recommended to create a shortcut with argument `autostart` to ensure that LocalSend knows that it was started automatically.

Example: `D:\LocalSend\localsend_app.exe autostart`

See #71 for more details.

jgaxmey:
> It is recommended to create a shortcut with argument `autostart` to ensure that LocalSend knows that it was started automatically.
> 
> Example: `D:\LocalSend\localsend_app.exe autostart`
> 
> See #71 for more details.

I added autostart in the shortcut, but it still doesn't start on boot.

SunshineboyZj:
The MSIX version can start automatically, but it cannot start silently and cannot add the autostart startup parameter like exe. Can you fix this problem，thank you

HalCyonNJ:
The latest MacOS version also can't start silently, But as myself I didn't restart the system frequently so. It will be nice if you guys add this little extra. Thanks for the app btw, It really helps me alot!

Mapaler:
> The MSIX version can start automatically, but it cannot start silently and cannot add the autostart startup parameter like exe. Can you fix this problem，thank you

I found a way to use it, but the cmd window pops up once.
https://github.com/localsend/localsend/issues/71#issuecomment-1725628652

Tienisto:
Fixed in #1467

Issue #82, opened 2023-01-31
Run in background on Linux


NamorNiradnug:
Add option (maybe command line) to run localsend in background (even without tray). This is implemented on Windows, but seems like it is not on Linux.

Tienisto:
Hi, this is an open bug in the dependency: https://github.com/leanflutter/tray_manager/issues/30

I don't know if running in background without tray is desirable. Users should be able to manually close this app without task manager

TheGB0077:
@Tienisto 
I did it... but the solution wasn't pretty at all 😩

![image](https://user-images.githubusercontent.com/39922116/233364281-bf00d970-6290-469d-839e-a2352a63ed6d.png)

tiger360:
> @Tienisto I did it... but the solution wasn't pretty at all weary
> 
> ![image](https://user-images.githubusercontent.com/39922116/233364281-bf00d970-6290-469d-839e-a2352a63ed6d.png)

can you explain how you did it please 😀 , I'm on PopOs

TheGB0077:
@tiger360 Is the most recent release not showing up the tray for you or are you questioning out of curiosity?

snydox:
I'm running Fedora 38 with GNOME 44 and it does run in the background. I also activated it to start when I turn on the computer. However, I'd like this app to start hidden rather than popping up whenever i start the computer.

![Screenshot from 2023-06-14 15-51-41](https://github.com/localsend/localsend/assets/13590875/a5fc7bfb-bd4d-405d-a661-b09842326868)

Tienisto:
This feature is only enabled on Windows but you can simulate it:

1. Find the settings file of LocalSend (it's a json): Add `"flutter.ls_auto_start_launch_minimized":true`.
2. Run the binary with the `autostart` parameter so it knows that it was started during boot.

snydox:
> This feature is only enabled on Windows but you can simulate it:
> 
>     1. Find the settings file of LocalSend (it's a json): Add `"flutter.ls_auto_start_launch_minimized":true`.
> 
>     2. Run the binary with the `autostart` parameter so it knows that it was started during boot.

Thanks. I'll give that a chance.

tiger360:
> This feature is only enabled on Windows but you can simulate it:
> 
> 1. Find the settings file of LocalSend (it's a json): Add `"flutter.ls_auto_start_launch_minimized":true`.
> 2. Run the binary with the `autostart` parameter so it knows that it was started during boot.

[...]

PR #548, opened 2023-06-24
feat: Enable autostart on Linux

TheGB0077:
Addressing autostart configuration on linux reported in #82 

And perhaps fixing #529

### DIFF
--- a/lib/util/native/autostart_helper.dart
+++ b/lib/util/native/autostart_helper.dart
@@ -2,6 +2,7 @@ import 'dart:io';
 
 import 'package:launch_at_startup/launch_at_startup.dart';
 import 'package:localsend_app/init.dart';
+import 'package:localsend_app/model/state/settings_state.dart';
 import 'package:logging/logging.dart';
 import 'package:package_info_plus/package_info_plus.dart';
 import 'package:url_launcher/url_launcher.dart';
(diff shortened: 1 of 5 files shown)


lib/util/native/autostart_helper.dart
```diff
@@ -2,39 +2,67 @@ import 'dart:io';
...
-Future<void> initAutoStartAndOpenSettings() async {
+Future<bool> initEnableAutoStartAndOpenSettings(SettingsState settings, [bool? isWindows]) async {
   try {
     // In case somebody don't use msix
     final packageInfo = await PackageInfo.fromPlatform();
     launchAtStartup.setup(
       appName: packageInfo.appName,
       appPath: Platform.resolvedExecutable,
-      args: [launchAtStartupArg],
+      args: [if (!settings.autoStartLaunchMinimized) launchAtStartupArg],
```
Tienisto, line 20: So you call `initDisableAutoStart` and `initEnableAutoStartAndOpenSettings` for the "launch minimized" setting only to update the `launchAtStartupArg` flag right?

TheGB0077: Yes, I ensured the deletion of the .desktop file / registry before writing it again

TheGB0077:
I found that this approach is not working with the .deb install. I will further debug to find out why.

lib/util/native/autostart_helper.dart
Tienisto: I don't know if this is needed as we already check the setting during startup: https://github.com/localsend/localsend/blob/d8babe05970066536162054511265e40013ad72b/lib/init.dart#L100

TheGB0077:
It looks like the library it's failing to identify the standard implementation that it should be using.

image


image

lib/util/native/autostart_helper.dart
TheGB0077: This check is only relevant for the initial boot of the App, but when writing the argument to the File / Registry it needs to follow the current settings.autoStartLaunchMinimized value, otherwise, it will always be written.

TheGB0077:
My bad. It was a mistake I introduced before pushing, and it went unnoticed 🫠

lib/util/native/autostart_helper.dart
Tienisto: > but when writing the argument to the File / Registry it needs to follow the current settings.autoStartLaunchMinimized value

Can you elaborate the other scenarios? Isn't `autoStartLaunchMinimized` only used for the initial boot of the app?
Assuming, `launchAtStartup` is `true`, `autoStartLaunchMinimized` is`false` and the `launchAtStartupArg` flag is enabled. The app should start normaly (not minimized) as expected as `autoStartLaunchMinimized` is `false` during init.

From my perspective, the flag `launchAtStartupArg` only tells LocalSend that the app was started during boot / login. The "launch minimized" feature is independent from this flag and only a feature built on top.

TheGB0077: Well, the changes made the `!persistenceService.isAutoStartLaunchMinimized()` check redundant indeed. Now that the autostart flag accurately reflects the user's choice. Tbh I forgot that this check existed until you pointed it out 😆, but I suppose that it does work both ways. The only difference is that the autostart argument will be written only when requested by the user.

Tienisto: Well, as long as it works :smile:

Tienisto:
Nice!

PR #597, opened 2023-08-01
deps: window_manager bump

TheGB0077:
Some important fixes were made: 

* On-close event handler is fixed
* setMinimumSize and setMaximumSize is fixed on macOS, although not yet on Linux.


Tienisto:
Thanks!

Issue #568, opened 2023-07-13
support file/image transfer by pasting files from the clipboard.


AlexYuan:
### Is your feature request related to a problem? Please describe.

I hope it will be possible to support file transfer by pasting files from the clipboard.

### Describe the solution you'd like

I copy files or images from the file manager, other software (such as Twitter App, Word, Photoshop) or web pages (such as twitter.com or t.cn) to the clipboard, and then paste them in LocalSend, which can be sent to other devices (it has already installed the same version of LocalSend).

![image](https://github.com/localsend/localsend/assets/1269376/47011fcd-4fbb-4963-ad25-197d95bdb699)
![image](https://github.com/localsend/localsend/assets/1269376/4a1c3cb0-6b69-4838-8eeb-2825427e6c9a)

Tienisto:
Thanks, your UI proposal looks good to me

Issue #210, opened 2023-02-27
Can i paste picture directly on localsend for windows


shui-dun:
### Is your feature request related to a problem? Please describe.

When I **take a screenshot** or copy an image on a web page, I must first find a way to save it as a file (for example, first paste the picture into Word, and then click "save as"), and then send it through localsend

### Describe the solution you'd like

I hope that localsend can directly read the contents of the clipboard (picture, better if it can be file in any format), and then I can send it.

### Describe alternatives you've considered

Add a paste button or allow to paste directly through ctrl+v

saccohuo:
+1

PR #674, opened 2023-09-08
Support file/image transfer by pasting files from the clipboard.

gitstart:
fixes https://github.com/localsend/localsend/issues/568
fixes https://github.com/localsend/localsend/issues/210

### PR Description
This PR add feature to send files that are copied to the clipboard 

### Demo video 
https://www.loom.com/share/929b2113cc4a487e9b78781c0e30facd

---
This code was written and reviewed by GitStart Community. Growing great engineers, one PR at a time.

### DIFF
--- a/assets/i18n/strings.i18n.json
+++ b/assets/i18n/strings.i18n.json
@@ -36,7 +36,8 @@
     "stop": "Stop",
     "save": "Save",
     "unchanged": "Unchanged",
-    "unknown": "Unknown"
+    "unknown": "Unknown",
+    "noItemInClipboard": "No item in Clipboard"
   },
   "receiveTab": {
     "title": "Receive",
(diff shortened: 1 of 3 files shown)


Tienisto:
Thank you! I will test this out in a few days.

Tienisto:
It works nice!

Do you know if this can be made more general? If I paste text, it does not work.

Thinking about this feature again, I think it would be simpler if a dialog is opened if the Flutter window gets refocused. Like "Do you want to use the clipboard contents? Cancel | Use Clipboard

Let me know if you can do that.

gitstart:
Hi @Tienisto thanks for reviewing PR.

We can try to make pasting text work but  do you think popup a dialog asking if user want to use clipboard content or not is a good user experience? If so then please let us know we will make changes accordingly.

Tienisto:
You are right. I will just merge this until we have a better idea.

Issue #461, opened 2023-05-10
Bug report: [500] Internal Server Error, can't send file to PC


daidaiworm:
### Describe the bug

almost the same issue, both version Localsend 1.7.0 and 1.9.1.
1 PC: win10
2 Android Smart Phones(Mi 10 Ultra & Mi Max 3)

Both Android Phones can send/receive file to/from each other smoothly.
PC can send file to 2 phones.
Android Phones can't send file to PC.

### To reproduce

2. Click on '...'
3. Scroll down to '...'

### Expected behavior

File should be sent as expected.

### Screenshots

![微信图片_20230510213449](https://github.com/localsend/localsend/assets/3594904/840b2373-1b51-4168-aadf-1f4a31536070)

### Smartphone (please complete the following information)

- Device: Mi 10 Ultra
- OS: MIUI 12.5.6 / Android 11
- Version: 1.7.0 / 1.9.1

Tienisto:
Do you see any reaction from the receiving device?

daidaiworm:
> Do you see any reaction from the receiving device?

No. Nothing happens on my PC.

daidaiworm:
I'm using 3rd party's anti-virus/firewall.
When encounting this weird issue, I turned off and quit those firewall, but failed all the same.

<img width="442" alt="zf_2023-05-12_181528" src="https://github.com/localsend/localsend/assets/3594904/9af08763-43d9-418f-88ef-19c4d8a14f0e">
<img width="225" alt="zf_2023-05-12_182020" src="https://github.com/localsend/localsend/assets/3594904/faf207dc-bc13-47b1-b39a-716d06cbfb96">

Tienisto:
Yes, error 500 indicates that the receiving device (Windows) did received the request but somehow it failed to handle it.

There might be an unusual setup which hasn't been handled yet.

Do you have a Downloads folder? Can you select another destination path in the settings?

daidaiworm:
I prefer portable version, so I downloaded 1.9.1_window.
When failed to receive, I tried different version on my pc
1.9.1_msix
1.9.0_window
1.9.0_msix

As for the path, I just changed from "D:\soft\LocalSend" to "D:\temp\LocalSend"
Issue remain same.
<img width="214" alt="zf_2023-05-12_234025" src="https://github.com/localsend/localsend/assets/3594904/558086a2-ae29-4009-9ffc-aab0bb7d4291">

daidaiworm:
[...]

PR #680, opened 2023-09-12
fix: Fallback to "$HOME/Downloads" when default downloads folder is unavailable

Sqbika:
Fixes: #461 

# Bug
When the system does not have the default Downloads folder defined (`XDG_DOWNLOAD_DIR` for Linux, `374DE290-123F-4565-9164-39C4925E467B` folderId in Windows) the application fails to send the files with an 500 response.

# Fix
The PR aims to fix it by manually defaulting to the Downloads folder by absolute path
On Linux by using the `$HOME` variable, which should be defined by default
On Windows by using the `HOMEPATH` variable

### DIFF
--- a/lib/util/native/get_destination_directory.dart
+++ b/lib/util/native/get_destination_directory.dart
@@ -17,10 +17,18 @@ Future<String> getDefaultDestinationDirectory() async {
     case TargetPlatform.windows:
     case TargetPlatform.fuchsia:
       var downloadDir = await path.getDownloadsDirectory();
-      if (defaultTargetPlatform == TargetPlatform.windows) {
-        downloadDir = Directory('${Platform.environment['HOMEPATH']}/Downloads');
-      } else {
-        downloadDir = Directory('${Platform.environment['HOME']}/Downloads');
+      if (downloadDir == null) {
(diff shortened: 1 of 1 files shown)


lib/util/native/get_destination_directory.dart
```diff
@@ -14,6 +16,12 @@ Future<String> getDefaultDestinationDirectory() async {
     case TargetPlatform.macOS:
     case TargetPlatform.windows:
     case TargetPlatform.fuchsia:
-      return (await path.getDownloadsDirectory())!.path.replaceAll('\\', '/');
+      var downloadDir = await path.getDownloadsDirectory();
+      if (defaultTargetPlatform == TargetPlatform.windows) {
```
Tienisto, line 20: Thank you for this fix. I think there is a missing null check if `path.getDownloadsDirectory()` returns not null.

Also please check if the new path `~/Downloads` exists with `Directory(path).existsSync()` and fallback to `~`.

Sqbika: Added them. The nullcheck got voided while the platform check happened

Tienisto:
Thank you!

Issue #345, opened 2023-03-30
Roadmap


Tienisto:
### v1.9.0
- [ ] Directory share (#121)
- [ ] Web share (#198)

### v2.0.0
- [ ] Protocol v2 (https://github.com/localsend/protocol/pull/5)
- [ ] UI redesign (#300)

### v2.1.0
- [ ] Favorites (#7)

Lynnde1Gr00t:
I recently started using the program and noticed the same error mentioned in issue 314, as I am using the Dutch language. I'm not sure when version 1.9 will be released, and it may take quite some time. Have you considered releasing an intermediate corrective release, such as 1.8.1, to address language-related issues? By the way, I am grateful for the application, it works great 😜

Tienisto:
@Lynnde1Gr00t Thanks. Releasing is quite time consuming as there are so many different platforms. A `1.8.1` is very unlikely

septs:
i think enforce encryption is REQUIRED

Tienisto:
@septs The encryption setting can be hidden under "advanced" but it is good for debugging. I don't think this should be removed.

Tienisto:
Furthermore, encryption has to be disabled for #198 because all browsers reject self signed certificates.

Tienisto:
@gidano @soya-daizu @sergd88 @Nixuge @esaracho @Neo1102 The `v1.9.0` will be released this week. I would be nice if you translate the missing strings 🙏

im-ichat:
> @gidano @soya-daizu @sergd88 @Nixuge @esaracho @Neo1102 The next `v1.9.0` will be released this week. It would be nice if you translate the missing strings 🙏

[i8n-localsend-NL-v1-9.txt](https://github.com/localsend/localsend/files/11291335/i8n-localsend-NL-v1-9.txt)


hopefully one day someone might help me figure out on how to propperly send this in.. but here is NL

Tienisto:
Thanks @im-ichat , I've added your translations.

Nixuge:
@im-ichat basically:

Go to https://github.com/localsend/localsend/ and click the "Fork" button (top right), you'll have a new repo created at https://github.com/im-ichat/localsend/

inside this repo, modify the file with the missing strings (in assets/i18n), you can do it directly from the web editor, then commit the change (basically save, bottom right in the web browser)

Once that's done, go to https://github.com/localsend/localsend/pulls and select "New Pull Request"

[...]

PR #698, opened 2023-09-21
The changes in translation are based on request [#671](https://github.com/localsend/localsend/pull/671)

sergd88:
I would also recommend adding an announcement about this minor translation change to the [Roadmap](https://github.com/localsend/localsend/issues/345), using it as an Announcement Board, as you've done before. This way, others who are helping or want to assist with the translation can see it and make corrections, thereby standardizing all translations at this point and making the application more cohesive.

Furthermore, at your discretion, it may be worth considering the possibility of including these changes in the **_missing_translations.json** file to encompass everyone.


Tienisto:
Thanks @sergd88,

I am too eager to release the next version as I don't have time for project. That's why in the last 2 releases (1.11.1 and 1.11.0), I have used automatic translation.

You could check out the diff if you are interested into the changes:
https://github.com/localsend/localsend/compare/v1.10.0...v1.11.1

I will add an announcement in the next version!

PR #701, opened 2023-09-21
Update Translations

graphemecluster:
Mainly `zh-HK`
Please, don't ever use machine translation for `zh-HK`. (See what I just commented on [this commit](https://github.com/localsend/localsend/commit/1fa4a513644c20443d141913005ed38102cdee89).) @-mention me a few days before each release instead.

### DIFF
--- a/assets/i18n/strings_zh-CN.i18n.json
+++ b/assets/i18n/strings_zh-CN.i18n.json
@@ -71,7 +71,7 @@
       "link": "通过链接分享"
     },
     "sendModeHelp": "解释",
-    "help": "请确保目标连接的是同一个 Wi-Fi 网络。",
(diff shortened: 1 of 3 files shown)


Tienisto:
Thank you @graphemecluster 

Similarly to #698:
I thought that automatic translation saves some time.
I used GPT-3.5 for that but it seems to have some issues...
Thank you for noticing this issue.
I will notifiy you and other translators before the next release!

graphemecluster:
Yes, unfortunately GPT-3.5 doesn't include enough corpus for Cantonese so it outputs Mandarin words frequently even if Cantonese is explicitly specified. GPT-4 works much better, but same as all translators, it can hardly take cultural differences into account (many expressions that would sound overly complicated have to be paraphrased into a simpler ones).

PR #707, opened 2023-09-27
Delete round icons on android

Jesse205:
Remove these round icons as they are never used (android:roundIcon does not appear in AndroidManifest.xml)


Tienisto:
Thank you

PR #716, opened 2023-10-02
translation of _missing_translations_iw.json

ShlomoCode:
I hope I did it right, I just manually combined the translations in the language file and deleted the _missing_translations_iw.json file

****
By the way, when is a version with the new translations expected to be released?


Tienisto:
Thank you. You don't need to delete the "missing translations" file because it will be generated by a tool anyways.

The next release might happen somewhere in October. I am not sure yet as I don't have much time for a release for now.

PR #718, opened 2023-10-03
Update make_config.yaml

gvenusleo:
In the KDE Plasma desktop environment, apps installed using `.deb` file are classified as `Lost & Found` . After adding `Utility` to `categories`, the app will be classified as `Utilities` .


Tienisto:
Thank you!

PR #722, opened 2023-10-07
i18n: Update missing translation for Ukrainian and Russian language (uk-UK; ru-Ru)

sergd88:
OLED has no suitable equivalents in either Ukrainian or Russian, as it is an abbreviation, so everyone uses OLED.


Tienisto:
Thank you!

PR #728, opened 2023-10-13
Enforce white background for logo - fixes #627

### LINKED ISSUE #627: The default icon color on the desktop need to be optimized?
### Describe the bug

The icon on windows desktop is not clear
image

### To reproduce

by default

### Expected behavior

More clear transparent icon

### Screenshots

_No response_

### Desktop (please complete the following information)

_No response_

### Smartphone (please complete the following information)

_No response_

### Additional context

_No response_

kra-mo:
This is to ensure that there is always an appropriate level of contrast.

The lack of contrast is an issue both on Windows and Linux with dark themes.

Here's an example:

image


Tienisto:
Thanks for your contribution. I think we should create a new asset instead specifically for the Windows icon. The images in the assets are used in multiple ways so the risk for unintended side effects are too high.

I will use your image and make the necessary changes for you.

PR #731, opened 2023-10-14
i18n: Update translations for Ukrainian and Russian (uk-UK; ru-Ru)

sergd88:


Tienisto:
Thank you. Would you check if "Clipboard" can be made shorter in Ukrainian and Russian? For example, in English I changed it to "Paste". I don't know if it's possible in Ukrainian and Russian.

Bildschirmfoto 2023-10-14 um 18 46 19

sergd88:
> Thank you. Would you check if "Clipboard" can be made shorter in Ukrainian and Russian? For example, in English I changed it to "Paste". I don't know if it's possible in Ukrainian and Russian.
> 
> Bildschirmfoto 2023-10-14 um 18 46 19

I don't know of a good equivalent in either Ukrainian or Russian that would describe the content and still look concise. For Russian, the noun "Вставка" is not a bad option, but it's not commonly used, so it's not the best choice. Therefore, I'll have to translate it as a verb. Hopefully, eight letters won't be too much.

sergd88:
I made changes. If there are any further translation issues, please let me know. I will try to fix them promptly.

sergd88:
> > Thank you. Would you check if "Clipboard" can be made shorter in Ukrainian and Russian? For example, in English I changed it to "Paste". I don't know if it's possible in Ukrainian and Russian.
> > Bildschirmfoto 2023-10-14 um 18 46 19

And one more question: when you press "Paste", does it immediately insert the last item from the clipboard, or does it open an additional dialog window/page?

Tienisto:
I directly inserts the item (any file or text)

sergd88:
> I directly inserts the item (any file or text)



> I directly inserts the item (any file or text)

Okay, understood, thank you. In that case, I'll keep with the current translation.

sergd88:
> I directly inserts the item (any file or text)

Okay, understood, thank you. In that case, I'll keep with the current translation.

sergd88:
> I directly inserts the item (any file or text)

And I would appreciate it if you could verify if this translation "Вставити/Вставить" fits without reducing the font size. Because in Ukrainian and in Russian it is 8 letters, and in the word "Clipboard", which you refused – 9, so the difference of 1 letter is not big. I might need to look for another option for the translation.

Tienisto:
If this is too difficult, then I might need to make the buttons bigger. This is only for desktop so it should be reasonable

sergd88:
> If this is too difficult, then I might need to make the buttons bigger. This is only for desktop so it should be reasonable

I don't know, maybe it's worth increasing the size. I see there are two more translations that already include this translation, in Persian and Hebrew. It might be fine with Persian because characters take up less space than the word "Clipboard", but with Hebrew, most likely, the same situation, in which word does not fit in size.

sergd88:
Additionally, I've noticed a not a good peculiarity for languages that use right-to-left script; their button frames duplicate (tried to demonstrate this in a screenshot). I'm not sure if it's worth opening a separate issue for this, as this issue most likely not have arisen recently, and I don't know if native speakers of these languages have reported it, so I'll leave it here.

Tienisto:
This right-to-left bug is a Flutter bug so there is nothing we can do on LocalSend side

PR #742, opened 2023-10-18
docs: use ** instead of ####

Jesse205:
`#` stands for headline, so the note should not be bolded with `#`, but with `**` instead.


Tienisto:
Thank you

PR #743, opened 2023-10-18
i18n: Update of the translation for the English language

sergd88:
I noticed several inaccuracies in the translation in the new paragraphs. I'll try to explain in more detail why, in my view, it would be better this way.

1. "No favorites devices yet" instead of "No favorites configured yet". As far as I understand, "Favorites" represents a list of devices you've selected, the most frequently used, and so on. It's the devices you add to the list, not configurations.

2. "Add" instead of "New". This follows from the first point. You're adding a device to the list, not creating a new one. Configurations are needed only for setting up the device itself.

3. "Delete from favorites" instead of "Delete favorite". You're deleting a device from the list of favorites, and, in my opinion, this sounds more logical than just delete favorite.

4. "Do you really want to delete from favorites" instead of "Do you really want to delete the favorite". This follows from the previous point.

5. "Add to favorites" instead of "Add favorite". As mentioned above, you are adding a device to the list, so the key idea is to "add to".

6. "Adjust" instead of "Edit favorite". As far as I understand, it specifically refers to configuration (IP Address, Port), so "adjust" sounds more logical than "edit."

7. "Alias" instead of "Name". It sounds more standardized and is consistent with previous translations, specifically with line 48, where "Alias" is already used.

With this translation, it is much easier to translate into other languages, at least in Ukrainian and Russian, and the translation sounds clearer. For now, I'm not submitting translations in Ukrainian and Russian, as I mentioned earlier, I strive to adhere to the most accurate translation in accordance with the original, so, most likely, depending on the approval of these changes, I will be guided when translating.


Tienisto:
Your explanation sounds reasonable. Thanks!

sergd88:
> Your explanation sounds reasonable. Thanks!

Oh, glad to hear that. Then I'll get started on the translations.

PR #753, opened 2023-10-20
Update strings_es-ES.i18n.json

kant:
* Issues in lines 91, 189, 237, 302, 314


Tienisto:
Thank you. Be aware that keys starting with `@` are comments. They will likely updated to English if I run the translation tool to apply your translations from #754

Issue #660, opened 2023-09-04
Confirm before deleting history


jasmyneokudo:
### Is your feature request related to a problem? Please describe.

When I click the delete history button, it deletes immediately without confirmation, 

### Describe the solution you'd like

I'll like to see a confirm dialog that asks me if I really want to delete my history

AshkanArabim:
I noticed the PR was merged—just wondering if there's anything else needed to close this issue?

Tienisto:
I've forgot to close this. Good catch!

PR #765, opened 2023-10-21
ADD: Alert while deleting  history

pantshaswat:
I added a dialog box when the user presses the Delete history button, asking for reconfirmation to delete history. I addressed issue #660. I just coded a box in receive_history_page.dart for now, can also make a widget for reconfirmation.

### DIFF
--- a/app/assets/i18n/strings.i18n.json
+++ b/app/assets/i18n/strings.i18n.json
@@ -144,6 +144,7 @@
     "title": "History",
     "openFolder": "Open folder",
     "deleteHistory": "Delete history",
+    "areYouSure": "Are you sure?",
     "empty": "The history is empty.",
     "entryActions": {
       "open": "Open file",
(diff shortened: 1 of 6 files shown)


Tienisto:
Thanks for your contribution. Since you probably have Flutter installed, you could edit the i18n file (https://github.com/localsend/localsend/blob/main/app/assets/i18n/strings.i18n.json), then after running `dart run slang`, you should have the new translations. Editing the generated file is not good because it will be overridden again

pantshaswat:
I just changed the i18n file and generated the file as you mentioned.

WebSnke:
Hi, I would like to translate the strings that will be added through this pull request. How should I approach that?

Tienisto:
@WebSnke Hi, the easiest way is just to wait until this PR is merged :)

You could also fork this PR (https://github.com/pantshaswat/localsend/tree/fix/660/confirm-delete-history) but there is a risk that something might be changed

WebSnke:
> @WebSnke Hi, the easiest way is just to wait until this PR is merged :)
> 
> You could also fork this PR (https://github.com/pantshaswat/localsend/tree/fix/660/confirm-delete-history) but there is a risk that something might be changed

Ok, thanks.

Tienisto:
Thanks. I refactored the code a little bit to keep the code style consistent

PR #767, opened 2023-10-22
Distribution: Update .deb / Add .rpm scripts

TheGB0077:
While this PR adds the RPM script, the best way to add it into the CI should be something to consider. Building it as is on a Debian will make it dependent on Debian specific libs, that cannot be found in Fedora land, so it may be necessary to build the app on a fedora container of storts to make it work out of the box properly.

### DIFF
--- a/app/linux/packaging/deb/make_config.yaml
+++ b/app/linux/packaging/deb/make_config.yaml
@@ -8,7 +8,7 @@ co_authors:
     email: gb.lima.dev@gmail.com
 priority: optional
 section: x11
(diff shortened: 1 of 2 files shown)


TheGB0077:
Here's a package I built on NobaraOS. I'm just not sure how compatible it is with Fedora 37 and below.

[localsend_app-1.11.135-linux.zip](https://github.com/localsend/localsend/files/13062220/localsend_app-1.11.135-linux.zip)

Tienisto:
Thank you!

PR #738, opened 2023-10-17
Update windows icon

Tienisto:
Follow-up for #728 

Closes #627

PR #864, opened 2023-11-03
Update Linux icon

kra-mo:
Follow-up for #728

This compliments #738

When the Flatpak is updated, it will need to be changed there too.

(Line 128 in org.localsend.localsend_app.yml)


Tienisto:
Hello, let's keep it as is fow now since not everybody like the circular white background (as seen in #738). You can open an issue if you really want an icon change

Issue #903, opened 2023-11-16
Add check if file is already selected


programmermager:
### Describe the bug

I noticed that in LocalSend, it is still possible to send the identical files multiple times. So, when I select a file, let's say file_a.jpg, and then select it again, the file is added to the queue again, resulting in the transmission of two files.

Is this the expected behavior of the LocalSend feature, or is it a bug @Tienisto ?

### To reproduce

1. Go to tab `Send`
2. Click on File or Media
3. Select a file or image or video
4. And then add the same file with pick previously

### Expected behavior

I feel that there should be validation within a session to prevent the reselection of files that have already been chosen, unless the sending session has changed.

### Screenshots

![Screenshot_20231116-162334](https://github.com/localsend/localsend/assets/7869219/46240f3e-80a4-4094-89e2-f14eeb2caffd)
![Screenshot_20231116-162341](https://github.com/localsend/localsend/assets/7869219/f7e6cbe0-99fb-4907-b4b4-fce2bc35b92f)

### Desktop (please complete the following information)

- OS : MacOs
- Version : MacOs Ventura 13.2.1

### Smartphone (please complete the following information)

- Device : Samsung A50s
- OS : Android
- Version  : Android 11

### Additional context

I tried in MacOs and Android. But i think this is will happens to all platform

Tienisto:
You are right, there is currently no check if the same file is selected again

programmermager:
> You are right, there is currently no check if the same file is selected again

But should that require validation? If yes, maybe I can help with that.

Tienisto:
Yes, a validation would be great

programmermager:
> Yes, a validation would be great

ok, i will try on it. do you think for text and copy paste should be add the duplicate validation too?

programmermager:
i already create the [PR](https://github.com/localsend/localsend/pull/904) for this enhancement, hope it will be merge. thx
@Tienisto

PR #904, opened 2023-11-16
add duplicate files validation when pick image or files or directory

programmermager:
linked to #903 

I try to add validation for already selected files from menu options file, media or folder.

# Before 
### MacOs
https://github.com/localsend/localsend/assets/7869219/843a0615-a9de-4d90-9c89-6ebc39572872

### Android
https://github.com/localsend/localsend/assets/7869219/7ef7ccd1-b6e3-4ec7-8794-b790eb104439

# After 
### MacOs
https://github.com/localsend/localsend/assets/7869219/1c17290e-a86c-44a0-bcf3-1c81348727fd

### Android
https://github.com/localsend/localsend/assets/7869219/ec4923ee-7900-4834-a5f4-0e9bf3687755

### DIFF
--- a/app/lib/provider/selection/selected_sending_files_provider.dart
+++ b/app/lib/provider/selection/selected_sending_files_provider.dart
@@ -2,12 +2,11 @@ import 'dart:convert' show utf8;
 import 'dart:io';
 import 'dart:typed_data';
 
-import 'package:flutter/material.dart';
-import 'package:image_picker/image_picker.dart';
 import 'package:localsend_app/model/cross_file.dart';
 import 'package:localsend_app/model/file_type.dart';
 import 'package:localsend_app/util/file_path_helper.dart';
 import 'package:localsend_app/util/native/cache_helper.dart';
+import 'package:localsend_app/util/native/cross_file_converters.dart';
 import 'package:logging/logging.dart';
 import 'package:path/path.dart' as p;
 import 'package:refena_flutter/refena_flutter.dart';
@@ -124,7 +123,7 @@ class AddFilesAction<T> extends AsyncReduxAction<SelectedSendingFilesNotifier, L
       //  https://github.com/fluttercandies/flutter_photo_manager/issues/589
 
       final crossFile = await converter(file);
-      final isAlreadySelect = state.any((element) => p.basename(element.path ?? '') == p.basename(crossFile.path ?? ''));
+      final isAlreadySelect = state.any((element) => element.isSameFile(otherFile: crossFile));
       if (!isAlreadySelect) {
         newFiles.add(crossFile);
       }
@@ -161,7 +160,7 @@ class AddDirectoryAction extends AsyncReduxAction<SelectedSendingFilesNotifier, 
           bytes: null,
         );
 
-        final isAlreadySelect = state.any((element) => p.basename(element.path ?? '') == p.basename(file.path ?? ''));
+        final isAlreadySelect = state.any((element) => element.isSameFile(otherFile: file));
         if (!isAlreadySelect) {
           newFiles.add(file);
         }
--- a/app/lib/provider/selection/selected_sending_files_provider.dart
+++ b/app/lib/provider/selection/selected_sending_files_provider.dart
@@ -2,6 +2,8 @@ import 'dart:convert' show utf8;
 import 'dart:io';
 import 'dart:typed_data';
 
+import 'package:flutter/material.dart';
+import 'package:image_picker/image_picker.dart';
 import 'package:localsend_app/model/cross_file.dart';
 import 'package:localsend_app/model/file_type.dart';
 import 'package:localsend_app/util/file_path_helper.dart';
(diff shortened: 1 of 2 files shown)


Tienisto:
Thank you. I think we need to check the absolute path and not the base path since users can add a directory (with its sub directories). There might be the case that you have 2 files of the same name but in different directories

programmermager:
> Thank you. I think we need to check the absolute path and not the base path since users can add a directory (with its sub directories). There might be the case where you have 2 files with the same name but in different directories

i see, the reason why i use the basename because the path from gallery and file is different although the file is same but it returns different path.

let me think the more effective way

programmermager:
@Tienisto I believe it's not possible to compare using an absolute path when the files come from the file picker and media picker. They are entirely different even if the files are the same. Below, I have attached the paths provided by the file picker and media picker.

pathimage

The green line represents the absolute path from the file picker, while the yellow one is from the media/gallery. It's the same file; the only thing that hasn't changed is the file name.

I would like to suggest continuing to use the basename for validation, but with additional comparison based on file size. Perhaps comparing only the file names could lead to errors, but if we compare based on both the file name and size parameters, I believe it would have high accuracy, unless there is one duplicated file.

WDYT?

WubTheGame:
> @Tienisto I believe it's not possible to compare using an absolute path when the files come from the file picker and media picker. They are entirely different even if the files are the same. Below, I have attached the paths provided by the file picker and media picker.
> 
> pathimage
> The green line represents the absolute path from the file picker, while the yellow one is from the media/gallery. It's the same file; the only thing that hasn't changed is the file name.
> 
> I would like to suggest continuing to use the basename for validation, but with additional comparison based on file size. Perhaps comparing only the file names could lead to errors, but if we compare based on both the file name and size parameters, I believe it would have high accuracy, unless there is one duplicated file.
> 
> WDYT?

Wouldn't comparing based on file hash be better? Perhaps a niche scenario, but you can (accidentally or not) name two unrelated files the same thing and if they manage to be the same size, that'd not be good. Though I guess it could also depend on how precise in terms of bytes the filesize check is (I don't know how those would work). Point is, I think there'd be a more effective way than a filesize check. In fact, a hash check would cover everything else about the file too. If checking for an absolute duplicate, I think you'd want a hash check.

Also wondering, why does Localsend copy files into a temporary directory of its own before sending them? What's the point of that? Why not just collect paths and grab from them at the moment of transfer? I can see that being a problem if you're low on storage.

Not a developer, so please pardon any misunderstandings.

programmermager:
> > @Tienisto I believe it's not possible to compare using an absolute path when the files come from the file picker and media picker. They are entirely different even if the files are the same. Below, I have attached the paths provided by the file picker and media picker.
> > pathimage
> > The green line represents the absolute path from the file picker, while the yellow one is from the media/gallery. It's the same file; the only thing that hasn't changed is the file name.
> > I would like to suggest continuing to use the basename for validation, but with additional comparison based on file size. Perhaps comparing only the file names could lead to errors, but if we compare based on both the file name and size parameters, I believe it would have high accuracy, unless there is one duplicated file.
> > WDYT?
> 
> Wouldn't comparing based on file hash be better? It'd would cover everything else about the file too. If checking for an absolute duplicate, I think that'd be the way to go.
> 
> Also wondering, why does Localsend copy files into a temporary directory of its own before sending them? What's the point of that? Why not just collect paths and grab from them at the moment of transfer? I can see that being a problem if you're low on storage.
> 
> Not a developer, so please pardon any misunderstandings.

I think it's a default in the file picker library used

Tienisto:
I think we should only exclude the same absolute file (when selected from the file picker) for now - which is still an improvement over the current version.

Only comparing the file size would cause false positives in rare cases which essentially a bug.

Calculating the hash would significantly increase the time during file selection (which is already quite slow sometimes).

programmermager:
btw the latest i tried to comparing the basename and file size, not only the size.

Are you saying it's acceptable for the same file to be included in the list of files to be sent when selected from the files and media menu?

Tienisto:
Yeah, comparing file name and file size just feels not right.

Ideally, file name and hash would be better but it's too slow.

To me, only applying the validation to "File" is enough if we use absolute paths.

That's just my opinion. I am open for other opinions.

programmermager:
I tried to perform a comparison using bytes. The problem wasn't particularly noticeable for image files. However, when selecting large files, such as videos and in large quantities, it takes too much time. So that options i believe we both doesnt agree to use that.

So, if you believe that a comparison based on name and size is not effective, it seems like the best option at the moment is an absolute path comparison. The drawback is that if the same file is retrieved using different methods, for example, via files and media, it will still be accepted and added to the list of files sent.

i will push new commit for comparation using absolute path

Tienisto (approved):

Tienisto:
Thank you. This should be fine for now.

programmermager:
My Pleasure, Hopefully I can contribute more here. 

I have some ideas that might be implemented in LocalSend

PR #914, opened 2023-11-19
i18n: Updated missing translations for Ukrainian and Russian languages (uk-UK; ru-Ru)

sergd88:


Tienisto:
Thank you. I have also added your name to [translators.dart](https://github.com/localsend/localsend/blob/main/app/lib/pages/about/translators.dart).

I guess that it is only your first name? In any case, you can update it if you want!

sergd88:
> Thank you. I have also added your name to [translators.dart](https://github.com/localsend/localsend/blob/main/app/lib/pages/about/translators.dart).
> 
> I guess that it is only your first name? In any case, you can update it if you want!

Thank you, I'll now add a last name to make it look more formal 😄

PR #916, opened 2023-11-19
Update Thai language latest

watchakorn-18k:


.gitignore
```diff
@@ -7,4 +7,4 @@
 
 .refena_inspector/
 
-/secrets
+/secrets
```
Tienisto, line 10: If think you removed the line break here.

Tienisto:
Thank you, besides the comment regarding the `.gitignore` file, you can also add your name to [translators.dart](https://github.com/localsend/localsend/blob/main/app/lib/pages/about/translators.dart) if you want to!

.gitignore
watchakorn-18k: > If think you removed the line break here.

Oh, I was going to ignore the flutter folder, but I decided to delete it so I removed it.

PR #919, opened 2023-11-20
i18n: Fix translation for Ukrainian language (uk-UK)

sergd88:
How long have I been using the app, but only today did I notice an inaccuracy in the translation in the most obvious and basic part 🤦‍♂️


Tienisto:
No worries!

Issue #861, opened 2023-11-03
Feature request: Trusted Device


trymeouteh:
### Is your feature request related to a problem? Please describe.

A list for devices that will allow Quick Save

### Describe the solution you'd like

A setting in a favorite device which will allow you to automatically accept file transfers from this device (Quick Save)

### Describe alternatives you've considered

None

sergd88:
Yes, such implementation makes sense, especially considering that now there's an option to add devices to the "Favorites" category. It would be something similar to how it's done in Google's Nearby Share. Though in that case, it relates to displaying the device on the network, whereas in our case, it's about Quick Save.

Then, instead of a toggle switch in the settings, we would have a dropdown list for Quick Save with the following options:

1. Disabled;
2. Enabled only for Favorite devices;
3. Enabled for all.

In combination with a excellent feature request [#860](https://github.com/localsend/localsend/issues/860), the "Favorite" function, in my opinion, would make much more sense, as currently, it seems not very useful.

Issue #250, opened 2023-03-07
[Feature request] Remember connected device.


hugeBlack:
### Is your feature request related to a problem? Please describe.

This app is so useful that I always use it to transfer files across platforms.
However, when I'm at school, devices cannot find each other even when they are in the same network, so I have to type in IP address of my devices every time to transfer files.

### Describe the solution you'd like

Add a "remembered devices" feature so that I can save IP address of all my devices since I seem to have a static nat IP address.
Whenever I want to send files to a device, I can simply choose the device in the remembered devices list without typing address all the time.

Tienisto:
Duplicate of #7

PR #945, opened 2023-11-28
feat: use favorite name in the device list

### LINKED ISSUE #860: Show favorite name instead of alias in the nearby device list

When you have several favorite devices saved, you will not know which device is which.

Please add a text field for favourite devices to label them. For example, if you want to saves your wife's device as a favorite you will set the name/label to "Wife's phone". This will help keep track of each favorite device and to know which one is which.

None

### Additional context

_No response_

Tienisto:
Closes #860

### DIFF
--- a/app/lib/model/state/server/receive_session_state.dart
+++ b/app/lib/model/state/server/receive_session_state.dart
@@ -13,6 +13,10 @@ class ReceiveSessionState with ReceiveSessionStateMappable {
   final String sessionId;
   final SessionStatus status;
   final Device sender;
+
+  // Might not be the same as sender.alias since it can be overridden as a favorite
+  final String senderAlias;
+
   final Map<String, ReceivingFile> files; // file id as key
   final int? startTime;
   final int? endTime;
@@ -24,6 +28,7 @@ class ReceiveSessionState with ReceiveSessionStateMappable {
     required this.sessionId,
     required this.status,
     required this.sender,
+    required this.senderAlias,
     required this.files,
     required this.startTime,
     required this.endTime,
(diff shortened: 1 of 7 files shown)


sergd88:
> Closes #860

Perfectly. Thank you. I would also recommend that you pay attention to [#861](https://github.com/localsend/localsend/issues/861). It sounds quite reasonable and should not pose any difficulties in implementation. Especially since similar topics have been raised many times before, and now with the introduction of Favorites, it is a rather logical development of the feature.

Tienisto:
Hi, unfortunately, there is currently no way to verify the authenticity of the fingerprint (a random string that is used for favorites).
Actually, I have added the "Favorites" feature because of #250 where the discovery is not working properly.

Flutter (the framework to build LocalSend) currently does not allow to retrieve the SSL certificate by the sending device. This is needed to check if the fingerprint (which is the hash of the SSL certificate) is actually valid.

I could implement #861 but anyone with technical expertise can fake the fingerprint (and therefore the device name).

sergd88:
> Hi, unfortunately, there is currently no way to verify the authenticity of the fingerprint (a random string that is used for favorites). Actually, I have added the "Favorites" feature because of #250 where the discovery is not working properly.
> 
> Flutter (the framework to build LocalSend) currently does not allow to retrieve the SSL certificate by the sending device. This is needed to check if the fingerprint (which is the hash of the SSL certificate) is actually valid.
> 
> I could implement #861 but anyone with technical expertise can fake the fingerprint (and therefore the device name).

Clearly. Sincerely grateful for the explanation. I thought something like a MAC address was used to identify the device and it would be possible to just compare the list. Sorry for any inconvenience. Thanks again.

PR #953, opened 2023-11-30
Update README.md

ShlomoCode:
`flutter` commands don't work in the root folder


ShlomoCode:
By the way, a warning appears -
`Deprecated. Use dart run instead.`

PR #955, opened 2023-11-30
Update zh-HK Translation

graphemecluster:
Plus fixing zh-CN translation


app/assets/i18n/strings_zh-CN.i18n.json
```diff
@@ -144,7 +144,7 @@
     },
     "noConnection": {
       "symptom": "双方设备均无法发现对方或者分享文件。",
-      "solution": "当问题发生在双方设备上时，请先确认双方设备处于同一 WiFi 网络内，且有相同的网络（端口、多播地址、加密选项）配置。若因 Wi‑Fi 不允许参与者间通信，那么请在路由器中关闭这个（如：AP 隔离）选项。"
+      "solution": "当问题发生在双方设备上时，请先确认双方设备处于同一 Wi‑Fi 网络内，且有相同的网络（端口、多播地址、加密选项）配置。若因 Wi‑Fi 不允许参与者间通信，那么请在路由器中关闭这个（如：AP 隔离）选项。"
```
graphemecluster, line 147: This change was made in #926. I don't know if it was intentional, but Wi‑Fi is a trademark that shouldn't be altered, and nevertheless “Wi‑Fi” in the second sentence was being kept.

graphemecluster: P.S. This is a non-breaking hyphen such that a line won't break between “Wi‑” and “Fi”. I made this change to the 3 Chinese locales in #701; I suggest making the same changes to other locales too.

PR #972, opened 2023-12-02
Added FR translations + small change to english strings

Nixuge:
Added FR translations for https://github.com/localsend/localsend/issues/345#issuecomment-1832988591

\+ While at it found out that there are 2 occurences of "open-source" in the main file, except one was written as "open-source" and the other one "open source", just made it so that both are now "open-source"


Tienisto:
Alright, thanks!

PR #1022, opened 2023-12-10
feat: remove local network permission check on iOS

Tienisto:


Allenxuxu:
Why do we need to remove this code? During my usage, I noticed that the app does not automatically check for Local Network permissions.

Tienisto:
I've noticed that it causes some crashes on iPhones and it seem that it works without this code.

PR #1109, opened 2024-01-29
Clean up some strings

Coopydood:
Simple changes to the English strings of the UI to be more professional.

E.g. wifi > Wi-Fi
Iphone > iPhone


Tienisto:
Thank you

Issue #1114, opened 2024-02-01
Feature request： should allow Anonymous Access the link share files on pc platfrom


MisterChangRay:
### Is your feature request related to a problem? Please describe.

for now, i need review every request and click button for allow those file be downloaded

i think, mostly user use the `share with link`,want share those no install client user,  and no one want take a long time to review those request,

### Describe the solution you'd like

if has a configuration to auto allow the request, it's better
or add password to the sharelink,  download user can be use password + link to download file

PR #1115, opened 2024-02-01
Bugfix 1114

MisterChangRay:
implements for #1114 

add allow anonymous Access the link share file, 

append global setting and  onece share settting
image

yes, you also can be setting on share link page, if global setting is off:
image

global settting will be auto override to the  share link page setting:
image

### DIFF
--- a/app/lib/provider/persistence_provider.dart
+++ b/app/lib/provider/persistence_provider.dart
@@ -67,7 +67,9 @@ const _sendMode = 'ls_send_mode';
 const _enableAnimations = 'ls_enable_animations';
 const _deviceType = 'ls_device_type';
 const _deviceModel = 'ls_device_model';
-const _anonymousAccess = 'ls_anonymous_access';
+const _anonymousAccessForGlobal = 'ls_anonymous_access_for_global';
+const _anonymousAccessForTemporary = 'ls_anonymous_access_for_temporary';
+
 
 final persistenceProvider = Provider<PersistenceService>((ref) {
   throw Exception('persistenceProvider not initialized');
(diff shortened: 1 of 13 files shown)


MisterChangRay:
it's test pass on windows10&win11&Ubuntu 22.04.3 LTS

app/lib/provider/persistence_provider.dart
```diff
@@ -112,6 +113,10 @@ class PersistenceService {
       await prefs.setInt(_version, 1);
     }
 
+    if (prefs.getInt(_anonymousAccess) == null) {
```
Tienisto, line 116: You don't need to initialize it here.
Initialization is only needed if you don't have a stable default value but I have seen that you have a default value (line 242)

app/lib/provider/persistence_provider.dart
```diff
@@ -233,6 +238,14 @@ class PersistenceService {
     await _prefs.setInt(_portKey, port);
   }
 
+  int getAnonymousAccess() {
+    return _prefs.getInt(_anonymousAccess) ?? 0;
```
Tienisto, line 242: If you intend to have 3 or more options, then an enum should be used (see `getSendMode`, `setSendMode` below).
An `int` is not type safe enough which could lead to bugs.

Tienisto (changes requested):

app/lib/provider/persistence_provider.dart
MisterChangRay: > Thinking again about this: You commented somewhere that `2` is only temporary. Then we only need to persist 2 states. In this case, a `bool` should be used here.

if use 2 states, only on/off can be use  ; 
in the share via link page, there is also a switch for temporay enablement, and it can't affect the global configuration.
so needs 3 states for this functional， globalOn/ globalOff/ onceOn
i will refactor code to enums later,

MisterChangRay: or use 2 configuration to done this , one use to global config, one use to temporary config

but i think int is better use to,  first Bit use to global config, and seceond Bit use to  temporary config

app/lib/provider/persistence_provider.dart
MisterChangRay: it's removed

app/lib/provider/persistence_provider.dart
MisterChangRay: > Thinking again about this: You commented somewhere that `2` is only temporary. Then we only need to persist 2 states. In this case, a `bool` should be used here.

i already refactor to bool

Tienisto:
Thank you

sergd88:
In my opinion, this isn't the parameter that most people will use because it's quite specific. Therefore, once again, in my opinion, it would be worth moving it to the "Advanced settings".

Tienisto:
Thanks @sergd88. I moved it to advanced settings now.

PR #1116, opened 2024-02-01
i18n/tr Translated missing turkish words

mertssmnoglu:
The last missing translation added to the i18n tr documentation. 
Looks like localsend is fully compatible with Turkish ( :tr: ) language.


Tienisto:
Thanks

PR #1172, opened 2024-02-28
headless cli

eugenechyrski:
just got  receive cli up and running over weekend, not ready to  be merged but a good start to continue  in that direction


Tienisto:
Thank you. The difficult part is to move as much logic as possible to the common package. I already added isolate handling for multi-threading in the future but I guess it's too much at once

PR #1177, opened 2024-02-29
feat: enable minify for android

Jesse205:
image

localsend has never enabled this feature, resulting in android packages being larger than almost all other platforms
image

### DIFF
--- a/app/android/app/build.gradle
+++ b/app/android/app/build.gradle
@@ -70,6 +70,8 @@ android {
     buildTypes {
         release {
             signingConfig signingConfigs.release
+            minifyEnabled true
+            shrinkResources true
         }
     }
 }


Tienisto:
Thank you, have you tested if it still works? I am surprised because Flutter should take the most amount of the binary.

Jesse205:
I tested it. It can send and receive pictures.
It can also send installed apps.

Jesse205:
In fact, the AndroidX library itself is very large, so enabling this switch will effectively reduce the size

Tienisto:
Unfortunately, I cannot reproduce the file size reduction with your config.

I have analyzed the released apk and it seems that `classes.dex` only takes less than 5 MB (uncompressed).

Bildschirmfoto 2024-03-01 um 01 18 54

Minification is still useful though.

Jesse205:
I'm very sorry for not verifying it. Since regular Android software behaves like this, I assumed the same for localsend.

Indeed, why🤔


I now find that the classes in dex are all confused, which proves that R8 is working. This switch should be turned on by default. But why does manually setting it have such a significant effect?🤔

Jesse205:
Well, I found out that he only kept the ARM 64 architecture, and I'm very sorry for the trouble.

Also I hope it's possible to release the different architectures separately, or provide the installer in apks (really a folksy installer format) format.

PR #1203, opened 2024-03-11
Serbian language translation strings

nebojsatomic:
Added Serbian translation file /app/assets/i18n/strings_rs.i18n.json


Mattoskid1:
Hi

Mattoskid1 (approved):

nebojsatomic:
Thanks @Tienisto  @Mattoskid1 :)
I will be keeping serbian branch on my fork for future improvements.
Also, in Serbian language there are two ways of writing, Latin, which is made available with this change, and Cyrillic, which could be added later perhaps, but for the time being Latin is enough.

Mattoskid1:
Oh okay you're welcome but I do have credit on my chaingpt account how can I withdraw it back to my wallet

PR #809, opened 2023-10-25
Update README.md

kushal34712:

Issue #1058, opened 2023-12-23
Bug report


nmummau:
### Describe the bug

I'm trying to compile LocalSend from the source code. I've forked, cloned, installed Flutter, etc. Errors pop up when I choose Linux or Chrome 

### To reproduce

1. Install [Flutter](https://flutter.dev).
2. Clone the LocalSend repository.
3. Run `cd app` to enter the app directory.
4. Run `flutter pub get` to download dependencies.
5. Run `flutter run` to start the app.
6. Get errors:

```
flutter run
Connected devices:
Linux (desktop) • linux  • linux-x64      • Ubuntu 22.04.3 LTS 6.2.0-39-generic
Chrome (web)    • chrome • web-javascript • Google Chrome 120.0.6099.129
[1]: Linux (linux)
[2]: Chrome (chrome)
Please choose one (or "q" to quit): 1

Launching lib/main.dart on Linux in debug mode...
CMake Error at flutter/ephemeral/.plugin_symlinks/system_tray/linux/CMakeLists.txt:29 (message):

  The `system_tray` package requires ayatana-appindicator3-0.1 or
  appindicator3-0.1.

Building Linux application...                                           
Exception: Unable to generate build files
```

or if you choose [2] Chrome:
```
flutter run
Connected devices:
Linux (desktop) • linux  • linux-x64      • Ubuntu 22.04.3 LTS 6.2.0-39-generic
Chrome (web)    • chrome • web-javascript • Google Chrome 120.0.6099.129
[1]: Linux (linux)
[2]: Chrome (chrome)
Please choose one (or "q" to quit): 2
Launching lib/main.dart on Chrome in debug mode...
../../../.pub-cache/hosted/pub.dev/extended_image_library-3.6.0/lib/src/network/network_image_web.dart:14:8: Error: Error when reading
'../../../snap/flutter/common/flutter/packages/flutter/lib/src/services/dom.dart': No such file or directory
import 'package:flutter/src/services/dom.dart';
       ^
../../../.pub-cache/hosted/pub.dev/photo_manager-2.8.1/lib/src/internal/image_provider.dart:71:5: Error: Type 'DecoderCallback' not found.
    DecoderCallback decode, // ignore: deprecated_member_use
    ^^^^^^^^^^^^^^^
../../../.pub-cache/hosted/pub.dev/photo_manager-2.8.1/lib/src/internal/image_provider.dart:96:5: Error: Type 'DecoderCallback' not found.
[...]

PR #1256, opened 2024-04-02
README: Reword and re-style "Getting Started" section

ix5:
Push users more heavily towards using `fvm` and get rid of nonsensical previous edits to this section.

Distro-provided flutter versions are almost certainly either too old or too recent for this roject (currently requiring Flutter SDK 3.13, yet e.g. Arch Linux provides an incompatible 3.19 in AUR), leading to unnecessary irritation and confusion when running subsequent `flutter` and `pub` commands.

---

The reason for me creating this PR was time wasted trying to make the `Flutter` version supplied by my distro work with LocalSend and finding out too late it relies on an outdated SDK version

> The issue may be caused ...

What even is that wording? Totally read right past that and assumed it was from a different section. I see https://github.com/localsend/localsend/pull/809 and https://github.com/localsend/localsend/pull/817 as well as https://github.com/localsend/localsend/pull/969 have actively worsened the readability of this section; it might be sensible to reject such garbage github-profile-inflating edits in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,15 +98,17 @@ For more information on the LocalSend Protocol, see the [documentation](https://
 
 To compile LocalSend from the source code, follow these steps:
 
-1. Install [Flutter](https://flutter.dev).
-2. Clone the LocalSend repository.
-3. Run `cd app` to enter the app directory.
-4. Run `flutter pub get` to download dependencies.
-5. Run `flutter run` to start the app.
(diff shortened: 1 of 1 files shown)


ix5:
See also https://github.com/localsend/localsend/issues/1058

As for ZH translations, perhaps better a native speaker adjusts them than me trying my luck with auto-translation services.

Tienisto:
Thanks!

PR #1282, opened 2024-04-22
Add Portuguese translations

hugok79:
I've added both **Portuguese** `pt` and **Portuguese (Portugal)** `pt_PT` lang codes because sometimes there are cases where Brazilian Portuguese `pt_BR` is loaded incorrectly instead of `pt`. So to prevent this from happening, it's best to add both lang codes. Both `pt` and `pt_PT` relate to Portuguese, the language spoken in Portugal.

Note: _missing_translations_pt.json/_missing_translations_pt_PT.json  files also have to be added or are generated automatically?


Tienisto:
Shouldn't `pt_PT` be enough to handle both Portuguese and Brazilian (which already has `pt_BR`)?

hugok79:
> Shouldn't `pt_PT` be enough to handle both Portuguese and Brazilian (which already has `pt_BR`)?

Maybe, `pt_PT` is enough to hande Portuguese (Portugal) then.

Although it's always necessary to have both `pt_PT` and `pt_BR` to avoid language confusions. Although Brazilians speak Portuguese, there are words they invented that don't belong in the Portuguese dictionary.

hugok79:
> Shouldn't `pt_PT` be enough to handle both Portuguese and Brazilian (which already has `pt_BR`)?

I deleted the `pt` and left only the `pt_PT` translation file

Issue #1274, opened 2024-04-16
Make share as link url larger


CubeOfCheese:
### Is your feature request related to a problem? Please describe.

I display the link on my screen in my classroom, but it's too small. QR code is not an option because the students are on desktop computers.

### Describe the solution you'd like

I want a button that makes the url display large on my screen. Similar to google classroom's class code UI and similar to the QR code UI that already exists in localsend, but for the text url instead of the QR code.

![Step-2-Click-a-Class-Card-01-4217695372](https://github.com/localsend/localsend/assets/16740469/396c00e0-8652-4cc5-88db-f6418dd6b3cc)
![maxresdefault-631120269](https://github.com/localsend/localsend/assets/16740469/43c42974-2e74-4e97-852b-0c2a0f1196c3)

### Describe alternatives you've considered

I zoom in with windows magnifier, but that's not elegant or ideal

Caesarovich:
This is an issue I also have. I'll try to make a PoC.

PR #1283, opened 2024-04-22
#1274 Make share larger

harriseldon:
I added a zoom button to the network share with a new zoom dialog. The dialog shows the url, but maximizes the size based on the screen width. I tested it manually on both a Linux desktop and Android tablet. If there are unit tests to write or other required changes before accepting, please let me know.


Tienisto:
Thank you.
I think the title "URL:" would fit better than "Zoom".
I played around a bit to avoid the hard coded font size but I couldn't find a better solution yet

harriseldon:
Thanks for the quick feedback. Sorry it took me so long for a simple text change. I updated the i18 json file and changed the title to URL per your request.

Tienisto:
Thank you

Issue #1301, opened 2024-04-29
Intermediate server, online list function


o2e:
### Is your feature request related to a problem? Please describe.

The receiver can't be searched for on the LAN, but you can find each other by ip address, all on the same network segment, under different routers.

### Describe the solution you'd like

Can you make a third party server to maintain the online user list, which can be deployed to a host on the LAN, and then each device fills in the ip of the server to get the online user list

PR #1304, opened 2024-04-30
add discovery timeout setting

o2e:
Improve cases where the timeout is too short to discover devices due to network quality by adding a timeout setting. #1301


Tienisto:
Just a note: Dart uses a lot of code generation, for example the translated strings are in [app/assets/i18n](https://github.com/localsend/localsend/tree/main/app/assets/i18n).

Tienisto:
Thank you!

Tienisto:
Thinking about this again, I think it is already at 2 seconds. `startupCheckAnotherInstance` is never used, `discovery` is using 2 seconds by default

PR #1305, opened 2024-04-30
feat: always receive trusted

yariefw:
Hello. I implemented a feature to enable quick save only for 'trusted' devices. Currently it just check if the sender is in the favorite device list and doesn't have it's own list (it's enough for me). Please take a look.

Related issue https://github.com/localsend/localsend/issues/861

### DIFF
--- a/app/assets/i18n/strings.i18n.json
+++ b/app/assets/i18n/strings.i18n.json
@@ -105,6 +105,11 @@
     "receive": {
       "title": "Receive",
       "quickSave": "@:general.quickSave",
+      "quickSaveType": {
+        "disabled": "Disabled",
+        "enabledForTrusted": "Only favorite devices",
+        "enabled": "Enabled for all"
+      },
       "autoFinish": "Auto Finish",
       "destination": "Destination",
       "downloads": "(Downloads)",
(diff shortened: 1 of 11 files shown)


Tienisto:
Currently, the HTTPS certificate is not checked so the fingerprint can be easily copied by an attacker.
This would create an illusion of safety as of now

yoiang:
@Tienisto I feel this is a lovely feature, what would it take for you to feel it is secure? Would it be enough if this PR was to include generation and secure storage of a private and public identifier and that was what was used to identify that it is the trusted source?

Tienisto:
We currently accept any certificate when sending files:

- send to server: https://github.com/localsend/localsend/blob/76aeef5a4ef027a4304333f1bce5c4016709278a/app/lib/provider/dio_provider.dart#L50

We currently accept any certificate when receiving files:
- receive from client: https://github.com/localsend/localsend/blob/76aeef5a4ef027a4304333f1bce5c4016709278a/app/lib/provider/network/server/server_provider.dart#L228

I would like to know if the certificate itself is valid, and compare it to the stored certificate.

Jkjjjaj (approved):

Noemi78 (approved):

Jkjjjaj (approved):

Noemi78 (approved):

Noemi78 (approved):

Noemi78 (approved):

Noemi78 (approved):

Noemi78 (approved):

PR #1331, opened 2024-05-13
i18n: fix typo in German translation

jakob30061:


jakob30061:
That was fast!
Thanks for developing and maintaining this great app

Tienisto:
Thanks for the fix!

Issue #1063, opened 2023-12-27
Rename files if the characters are not supported on the receiving operating system


alexmozaidze:
### Describe the bug

When trying to send a file with characters permitted on sender's OS, but illegal on receiver's OS, it errors with:
- Sender: `[500] Could not save file`
- Receiver: `PathAccessException: Cannot open file, path = '/bla/bla/my:cool:video.mp4' (OS Error: Operation not permitted, errno = 1)`

The error does not occur if `my:cool:video.mp4` is renamed to `my_cool_video.mp4`.

### To reproduce

1. Create a file with some funky characters that are illegal on the receiver's OS

```
my:cool:file.txt
```

2. Send this file to the receiver
3. Gaze upon the error messages on both devices

### Expected behavior

It should probably replace illegal characters with underscores, since most operating systems allow them in filenames. It would also be nice if the user was notified beforehand when the file will be renamed on the receiver's end.

### Desktop (please complete the following information)

- OS: Void Linux (ext4 filesystem)
- LocalSend Version: 1.13.1

> [!NOTE]\
> LocalSend was installed and ran using flatpak.

### Smartphone (please complete the following information)

- Device: Xiaomi Redmi 10C
- OS: Android 13
- LocalSend version: 1.13.1

Caesarovich:
I've made a PoC using the [`sanitize_filename`](https://pub.dev/packages/sanitize_filename) package and it fixes the issue. 
But before making a PR I would like to discuss further on the implementation. I have two main concerns:

1. The `sanitize_filename` library doesn't allow for granular sanitization. Which means that a filename that could be valid on a receiving Linux device, would still have the "Windows illegal characters" stripped from the filename. 
One might justify it as a *feature* because "This file's name is now properly named/It shouldn't have such characters anyway". But if we do decide that illegal characters should only be stripped if illegal on the target device, then I think it would be pretty easy to create a more granular version of `sanitize_filename`. 

2. The current UI (receive_tab) only displays the original filename. It would need to be updated to show the actual filename saved to the device. Or better there could be an indication that it is different from the original filename.


-> This issue is also related to this one #1180

Tienisto:
Thank you @Caesarovich .
Since you can change the file name as a recipient already, we can reuse this feature to change it beforehand.
I think that your library is already quite good. A flag that the recipient is a Windows device is better but not important

Caesarovich:
[...]

Issue #1180, opened 2024-03-01
Error when tansferring files with special characters.


tuqueque:
### Describe the bug

When I try to Drag&Drop some file with special characters  (ñ, |, :, 🙂) into LocalSend, it doesn't do anything... and even worse, if I Drag&Drop a folder with files inside that have special characters, the transfer takes place, but at the end of the file transfer, throws an error.

### To reproduce

Either have a file with special characters or a folder with files inside that have especial characters and try to transfer it with LocalSend.

### Expected behavior

Transfer the file!

### Desktop (please complete the following information)

OS: Manjaro (Linux).
LocalSend version: 1.14.0

tuqueque:
The target OS is Windows, yes... Although, I don't see why when I drag and drop a file with special characters into LocalSend, it doesn't do anything but if I place the file into a folder (that doesn't have special characters) the Drag&Drop function does work... and also, I really doubt that I'll have any problem if I send files with special characters through other methods like USB drive or even croc ( https://github.com/schollz/croc )... I can do the tests if you want.

So... is this fixable?

Tienisto:
I just sent a file named `ñ.png` and `🙂.png` and it did worked. But when adding a `|` it didn't because Windows doesn't support this character.

I don't know how other solutions work but I can imagine that they replace unsupported characters.

tuqueque:
Does it support Drag&Drop into LocalSend? (dragging and dropping files with special characters into LocalSend's window).

CorvinYu:
i meet the question too.
![image](https://github.com/localsend/localsend/assets/138097462/ce0062c9-12c8-4451-a15f-f64eb535569c)

whishing sort.

Tienisto:
Fixed in #1336

Issue #1101, opened 2024-01-21
[Android] [Bug report] FAT32-invalid characters result in a failed transfer


Pixadus:
### Describe the bug

When attempting to send a file that contains characters considered invalid by FAT32 including `* / : < > ? \ | + ; = []`, the file will be unable to send with error 500. 

### To reproduce

1. Create a file (such as <hello>.txt or Part1:TheTest.txt)
2. Send the file to an Android device running Android 13
3. Transfer will "finish", but then return error 500 afterward.

### Expected behavior

It was not immediately apparent that the file naming was an issue - this is more a problem with Android providing vague error messages when invalid file names are supplied. 

One idea is that LocalSend's Android version could sanitize file names of FAT32 invalid characters. 

### Desktop (please complete the following information)

- Host OS: Linux

### Smartphone (please complete the following information)

- Version: 1.13.1
- Android 13

Tienisto:
Fixed in #1336

PR #1336, opened 2024-05-15
Add filename sanitization

Caesarovich:
-> Fixes issue #1063 #1180 #1101  (And possibly more but I couldn't find)

## Description
As discussed in issue #1063 there currently is an error if a file is sent on a device with a different system of the sender. This is due to different character limitations in files' names depending on Operating systems and partitions. For example:

On my Linux computer I can create the file `<cool>.txt` and send it to my other linux computer without issue. But if I try to send it to my Windows computer or to my Android phone, LocalSend will fail to save the file and fail the transfer with error code 500. This is because `<>` are **illegal characters** on Windows and FAT32 devices (Android).

This PR fixes that issue by providing a sanitization step before saving a file. Like here:

```dart
Future<String> digestFilePathAndPrepareDirectory({required String parentDirectory, required String fileName}) async {
  final actualFileName = legalizeFilename(p.basename(fileName), os: Platform.operatingSystem); // <- here
  final fileNameParts = p.split(fileName);
  .....
```

The sanitization is handled by the [legalize](https://pub.dev/packages/legalize) library that I made for the occasion. So if any issues like this open up, they can be addressed upstream [here](https://github.com/Caesarovich/legalize/issues) instead. I'd be glad if users could test it across the many different OS they possess, I couldn't test it directly on all the possible different systems.

## Changes

- Now includes the `legalize: 1.2.1` dependency
- Filename sanitization in the `digestFilePathAndPrepareDirectory` function
- Filename check for the `FileNameInputDialog` and `QuickActionsDialog` widgets

### DIFF
--- a/app/assets/i18n/strings.i18n.json
+++ b/app/assets/i18n/strings.i18n.json
@@ -424,6 +424,10 @@
       "link": "Recipients who do not have LocalSend installed can download the selected files by opening the link in their browser."
     }
   },
+  "sanitization": {
+    "empty": "Filename cannot be empty",
+    "invalid": "Filename contains invalid characters"
+  },
   "tray": {
     "@info": "Apple Guidelines are very strict about the 'close' wording.",
     "open": "@:general.open",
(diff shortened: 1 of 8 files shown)


Caesarovich:
I've lowered the SDK requirements :)

Tienisto:
Thank you!

PR #1358, opened 2024-05-22
Fastlane: Translate into PT-BR

lucasmz-dev:
This PR:

* Translates short and full descriptions
* Adds translated screenshots*

*Screenshots are in dark mode, in a different aspect ratio, and are .png instead of .jpg. If this is not ideal, let me know and I'll remove them, and the regular English ones should be used instead, or create new ones.

`short_description` should be the right size for F-Droid IIRC. It uses only 66 characters.


Tienisto:
Those screenshots are crawled by the F-Droid bot. The publishing process to Google Play Store is currently not automated.
Let's see if it uses your screenshots in the next update :)

PR #1370, opened 2024-05-24
Create _missing_translations_hi.json

iharshsavaliya:


Tienisto:
Your PR does not make any sense. Why do you specify missing translations instead of translating them right away? This file is generated automatically, no need add manually

PR #1397, opened 2024-06-06
added mica to window and titlebar of app

bharathajjarapu:
Hello, I have causally changed theming of app to support mica so it looks good, I used `flutter_acrylic` package to support mica for titlebar and background of app, also i have formatted the `theme.dart`. Here's is how it looks

https://github.com/localsend/localsend/assets/106143429/865e3e54-0030-4bb4-bde2-e62742fbedaf

### DIFF
--- a/app/lib/main.dart
+++ b/app/lib/main.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:flutter_acrylic/flutter_acrylic.dart';
 import 'package:flutter_localizations/flutter_localizations.dart';
 import 'package:localsend_app/gen/strings.g.dart';
 import 'package:localsend_app/init.dart';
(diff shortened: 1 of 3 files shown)


Tienisto:
Thank you for the pull request!
Sadly, it is planned to support Windows 7 for at least one release (1.15.0).
Also, the mica effect seem to only work on Windows 11 but not on Windows 10, we will lose another chunk of users.

bharathajjarapu:
> Thank you for the pull request! Sadly, it is planned to support Windows 7 for at least one release (1.15.0). Also, the mica effect seem to only work on Windows 11 but not on Windows 10, we will lose another chunk of users.

thanks for a reply and app, i use the app everyday so had some time to have mica in it. I understand the app need to support windows 7 and appreciate your concern, again thanks for the app.

JustAdreamerFL:
Can there be a toggle for people that have Win11 to enable Mica theme or the default theme?

PR #1471, opened 2024-07-10
feat(macos): handle files that were dropped into the app icon

### LINKED ISSUE #1396: Better macOS interoperability - drag and drop to icon

The macOS app does not behave like the user expects a macOS to behave

- drop a file onto the app icon or Cmd+Tab switcher

Expected
- app should open file

Actual
- error "The document “filename.ext” could not be opened. LocalSend cannot open files in the “Long Description of File Type” format."

have to open the app then add file

### Additional context

- related #1393

Tienisto:
Closes #1396


ShlomoCode:
@Tienisto Is there any setting that needs to be enabled to use this feature?
I use macOS 14.5 and localsend v1.15.0 or v1.15.3, I drag an mp3 file to the menubar icon of LocalSend and nothing happens...
This is my settings:
CleanShot 2024-08-08 at 07 18 39@2x

ShlomoCode:
Nor does the macos darg symbol (the green plus sign) appear when I drag the file over the symbol

ShlomoCode:
Oh, my bad. Now I understand that this is the Dock icon and not the menu bar icon...

Can I suggest native code to integrate with the **menu bar icon**?
I think it will be much more useful than the Dock icon, because it allows files to be sent easily even when the application was running "in the background" - in the menu bar only.

I don't know how to integrate it with Flutter, but it works flawlessly as a native macOS app:
```
import SwiftUI
import Cocoa

@main
struct TestDragDropMenuBarApp: App {
    @NSApplicationDelegateAdaptor(AppDelegate.self) var appDelegate
    
    var body: some Scene {
        Settings {
            EmptyView()
        }
    }
}

class AppDelegate: NSObject, NSApplicationDelegate {
    var statusItem: NSStatusItem?
    
    func applicationDidFinishLaunching(_ notification: Notification) {
        statusItem = NSStatusBar.system.statusItem(withLength: NSStatusItem.squareLength)
        
        if let button = statusItem?.button {
            button.image = NSImage(systemSymbolName: "globe", accessibilityDescription: "Hello World")
            
            let dragView = DraggableView(frame: button.bounds)
            button.addSubview(dragView)
            
            dragView.translatesAutoresizingMaskIntoConstraints = false
            NSLayoutConstraint.activate([
                dragView.topAnchor.constraint(equalTo: button.topAnchor),
                dragView.leadingAnchor.constraint(equalTo: button.leadingAnchor),
                dragView.trailingAnchor.constraint(equalTo: button.trailingAnchor),
                dragView.bottomAnchor.constraint(equalTo: button.bottomAnchor)
            ])
        }
    }
}

class DraggableView: NSView {
    override init(frame frameRect: NSRect) {
        super.init(frame: frameRect)
        setup()
    }
    
    required init?(coder: NSCoder) {
        super.init(coder: coder)
        setup()
    }
    
    private func setup() {
        registerForDraggedTypes([NSPasteboard.PasteboardType.fileURL, NSPasteboard.PasteboardType.URL, NSPasteboard.PasteboardType.string])
    }
    
    override func draggingEntered(_ sender: NSDraggingInfo) -> NSDragOperation {
        return .copy
    }
    
    override func performDragOperation(_ sender: NSDraggingInfo) -> Bool {
        let pasteboard = sender.draggingPasteboard
        
        if let urls = pasteboard.readObjects(forClasses: [NSURL.self], options: nil) as? [URL] {
            for url in urls {
                print(url.path)
            }
            return true
        }
        
        if let strings = pasteboard.readObjects(forClasses: [NSString.self], options: nil) as? [String] {
            for string in strings {
                print(string)
            }
            return true
        }
        
        return false
    }
}
```
It just prints the path, of course you can change that to the real action (broadcast the path to the dart part of the app)

And here it is as a runnable Xcode project so that you can get an impression:
[TestDragDropMenuBar.zip](https://github.com/user-attachments/files/16537274/TestDragDropMenuBar.zip)

ShlomoCode:
I deleted my comments, opened a separate issue for this - https://github.com/localsend/localsend/issues/1615

PR #1487, opened 2024-07-11
Update zh-* Translations

graphemecluster:


app/assets/i18n/_missing_translations_zh_HK.json
```diff
@@ -5,36 +5,36 @@
...
     },
     "network": {
-      "discoveryTimeout": "Discovery Timeout"
+      "discoveryTimeout": "裝置搜尋逾時"
     }
   },
   "receiveHistoryPage": {
     "entryActions": {
-      "showInFolder": "Show in folder"
+      "showInFolder": "喺檔案瀏覽器顯示"
```
graphemecluster, line 16: I assume you mean “Show in _File Explorer_”

app/assets/i18n/_missing_translations_zh_HK.json
```diff
@@ -5,36 +5,36 @@
   ],
   "settingsTab": {
     "general": {
-      "showInContextMenu": "Show LocalSend in context menu"
+      "showInContextMenu": "喺檔案功能表嘅「傳送到」項目顯示 LocalSend"
```
graphemecluster, line 8: This should be clarified in other languages too, to:
Show LocalSend in the “Send to” file context menu

app/assets/i18n/_missing_translations_zh_HK.json
Tienisto: I didn't want to mention "explorer" because this is only correct on Windows and not on macOS for example

graphemecluster: Ah, I mean file manager/broswer (the latter is what I translated into)

PR #1488, opened 2024-07-11
Update _missing_translations_fr.json

Nixuge:
Saw that there's a release soon in https://github.com/localsend/localsend/issues/345#issuecomment-2221875935

Translation differences:
- Unsure if "menu contextuel" is the best fit as i've pretty much never seen it used like that, but can't find anything better for now.
- Changed meaning from "show in folder" to "open the folder" since "show in folder" is pretty unnatural in french
- Added "code" before pin because it sounds better
- changed meaning from "too many attemps" to "too many bad attempts"

### DIFF
--- a/app/assets/i18n/_missing_translations_fr.json
+++ b/app/assets/i18n/_missing_translations_fr.json
@@ -5,26 +5,26 @@
   ],
   "settingsTab": {
     "general": {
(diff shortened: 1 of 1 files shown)


Tienisto:
Thanks

Issue #907, opened 2023-11-17
Sending too many files froze the Android app


szabolcsdombi:
### Describe the bug

I have successfully used the android app to send 1k images over to my Windows computer.
I have selected the DCIM folder with 11k files in it. After a few seconds the files count and the total size of 85GB was shown (this supprized me as I cannot transfer this folder with an USB due to it hangs forever to show the files).
The app clearly froze, I have reset the app (force stop + delete data + cache) and retried.
It clearly hangs.
I was waiting for 10 minutes to see if it resolves, but not.
To define the frozen app:
Nothing happens if I touch any buttons including the Receive, or device list refresh,
I have tried to first press the device I wanted the files to be sent to.
I even tried long pressing emulating a simple press for a slow ui.

### To reproduce

1. Open the android app
2. Select aa images folder to send with 10k+ files with 80GB+ content.
3. Wait a few seconds to see the selection ready on the screen.
4. Try to press any other button.

### Expected behavior

To find some loading screen, or at least some "busy doing stuff" label, or some moving components to visualize it is not frozen.
It might be busy sending the metadata over.

### Smartphone (please complete the following information)

Pixel 6
Android 14
v1.8.0

programmermager:
Wow its a huge files to send over wifi

szabolcsdombi:
I could actually tether with the USB cable getting a wired connection. I do not mind the speed so I am using wifi. All types of the USB file transfers seems to fail at this file count.
Would it be possible to get past the frozen screen?
Do you have an idea why it hangs?
I can attach an adb if that would help. I do not know what to run, I use adb mostly to debug chrome.

sergd88:
> I could actually tether with the USB cable getting a wired connection. I do not mind the speed so I am using wifi. All types of the USB file transfers seems to fail at this file count. Would it be possible to get past the frozen screen? Do you have an idea why it hangs? I can attach an adb if that would help. I do not know what to run, I use adb mostly to debug chrome.

To transfer hundreds of gigabytes, I personally use Syncthing. However, it's a synchronization application, so there are some drawbacks. You can temporarily synchronize the entire contents of the DCIM folder with your computer through it, so you don't have to copy all the necessary content to another folder strictly for synchronization. Because additional copying requires free memory for this content, and it also wears out the lifespan of your storage device.

Also, of course, it supports resumption of copying/synchronization, so in case of a connection loss, you can resume copying/synchronization without the need to start everything from the beginning.

[...]

Issue #829, opened 2023-10-26
Bug report: After the transmission is completed, the selection is not cleared (Android)


sergd88:
### Describe the bug

The issue has resurfaced again [#321](https://github.com/localsend/localsend/issues/321), [#195](https://github.com/localsend/localsend/issues/195), and for some reason, now when clearing the app's data and cache, the application settings are retained. This implies that the configuration files are located in a different place, which is also quite problematic because it makes it impossible to reset all of the application settings to their default state through standard methods.

The issue is once again that, for some reason, the transmission page itself is not displayed (see screenshot), and the transmission is carried out directly from the "Send" tab.

### To reproduce

...

### Expected behavior

...

### Screenshots

![1](https://github.com/localsend/localsend/assets/125032334/3e36b216-77c4-435f-a56b-09349090aeeb)

### Smartphone (please complete the following information)

I tried several different devices with various firmware and versions of Android.

Tienisto:
Is it single recipient mode or multiple recipients mode? Somehow it works on my android phone

sergd88:
> Is it single recipient mode or multiple recipients mode? Somehow it works on my android phone

Single.

sergd88:
Somewhat everything is very strange. Background story. I updated from version 1.11.1 to version 1.12.0 through Google Play. Because sometimes updates in Google Play unfolded quite slowly, I would delete the version from Google Play and install the version from GitHub. As far as I understand, they have different digital signatures, so they are not compatible with each other in automatic updates through Google Play. Over time, when I saw that the new version was already available on Google Play, I would delete the version I had installed from GitHub and install the version from Google Play. In all these manipulations, I always cleared the app's cache and data so that no unpleasant situations would occur later, like the one I have now 😅. This time the update came quickly, so I didn't perform such manipulations, I simply updated the app through Google Play.

[...]

Issue #321, opened 2023-03-24
Feature request: Quite a strange behavior with the selected file after transferring it in Single recipient mode


sergd88:
### Is your feature request related to a problem? Please describe.

I also noticed that when sending to a Single recipient, the file remains selected after the transfer. I saw that this was intended, but it only works on the phone. When transferring through the app on Windows, the file is deselected after the transfer.

This is a somewhat strange default behavior that I couldn't understand. Why leave the file selected after transferring it to the phone? If it's to transfer it again to another device, then that's what the Multiple recipients mode is designed for. The most common use case is probably still transferring a file to a single recipient. Therefore, it would be much more convenient to deselect the file after the transfer, because if you need to transfer another file that you forgot about, you'll have to manually deselect the previous file to avoid transferring it again. I personally ran into this situation today.

### Describe the solution you'd like

-

Tienisto:
What kind of files did you sent? This may be related to #293

sergd88:
> 

In my case, it's media files, namely images.

Tienisto:
What LocalSend version do you use on Windows?

sergd88:
> What LocalSend version do you use on Windows?

The latest version of the program is 1.8.0 (27) and it is the same on both the phone and the computer.

But the issue with selection occurs not when send from computer to phone, but the other way around, from phone to computer.

If send from the computer to the phone, the selection disappears after the transfer, but if from the phone to the computer, it does not.

sergd88:
> What LocalSend version do you use on Windows?

Tomorrow I'll try to send file it from another Android device with a different OS and version. Maybe it's not affecting all of them.

sergd88:
> What LocalSend version do you use on Windows?

Tomorrow I'll try to send file it from another Android device with a different OS and Android version. Maybe it's not affecting all of them.

sergd88:
> What LocalSend version do you use on Windows?

I confirm that this is a local problem, specifically some kind of glitch, as everything works fine when send from another device.

It can be resolved by simply reinstalling the application on the device and clearing its data and cache.

And the selection not disappears because for some reason the device connection page itself was not displayed (see screenshot), and the transfer was made directly from the Send tab.

Therefore, you can close this issue.

![2023-03-25_11-08](https://user-images.githubusercontent.com/125032334/227708671-e900c0c6-e582-4e0b-b6ab-61aa942e02fb.png)

Tienisto:
Thanks!

Issue #220, opened 2023-03-01
Microsoft Store release


Tienisto:
This app was on the Microsoft Store in the past but on 27th February 2023, they removed the app without giving an exact reason.

This was their response after I asked them why they removed the app (and also the developer account!):

> Hello,
> Microsoft removed your developer account from the Microsoft Store based on good faith and reasonable evidence of activities that violate the terms of the account agreement.
> The account agreement allows Microsoft to remove a Microsoft Store Account for abuse of the marketplace or failure to keep your account in good standing. Termination of your account results in the removal of associated application(s) from the store. Also, the account agreement allows Microsoft to remove an application from availability in the store for breach of the agreement or for any other reason.
> 
> We have determined that the removal of your account and published app(s) were warranted based on abusive and/or inappropriate activities. Please note that our decision is final, and when one account is in violation of the Agreement, we may also remove any related accounts.
> 
> Thank you,
> Microsoft Store Team

The follow-up response after I asked them again:

> Our reply in this email provides the grounds for the action, which are included in the terms of the Application Developer Agreement. This is the full extent of information that we can provide, which is consistent with the Agreement.

The developer account only contained LocalSend and no other app. So LocalSend is most likely causing this issue.

techgamers2:
please try to add windows store version its easier to install and update tht way ... and thank u for ur work really like the app i switched from sendanywhere to ur app :)

Tienisto:
I just contacted Microsoft again, same response.

Tienisto:
This is weird. LocalSend is available on nearly every market place like Play Store, App Store and even Amazon Store. But Microsoft doesn't seem to like this app.

rollingmoai:
We don't even know the reason why. How are we supposed to solve the issue?

rollingmoai:
Did you try to reach out via a new e-mail to reportapp@microsoft.com?

Tienisto:
Yes. This is the email address that I used.

Sadly, we can only hope that the Microsoft team changes their mind.

I could not compile the `msix` installer yet because I am still waiting for an approval of an alternative certificate authority. Installing and updating via `msix` is at least not that bad.

NULL204:
In fact, I would prefer an EXE program pair installer rather than going through the Microsoft App Store, so why not consider making an EXE installer?

NULL204:
I think Clash For Windows is a good example if we are thinking about the automatic update function

Tienisto:
[...]

PR #1492, opened 2024-07-12
Updating missing translations, minor fixes, and standardizing translations for Ukrainian, Russian, and English languages

sergd88:
This is not related to translations;
I'd sincerely appreciate it if you could additionally publish versions of the Android app in your GitHub repository with a Google Play signature. This would allow quicker testing of new versions of the app, avoiding the wait time for Google's moderation, and also minimize issues with Google "rejecting updates", as happened with version 1.14.0, which likely resulted in fewer users being able to try it.


Tienisto:
Thank you. I believe Google won't accept this update also because the problem isn't solved yet. I could upload an additional APK that is signed by Google but users might be confused (as they don't know the difference between those versions)

sergd88:
> Thank you. I believe Google won't accept this update also because the problem isn't solved yet. I could upload an additional APK that is signed by Google but users might be confused (as they don't know the difference between those versions)

Yes, I understand that this might be somewhat confusing. However, to minimize confusion, I think the version signed with the Google Play signature should only be placed in the repository without a download link on the website. Also, naming it using a key, as you have done before for various fixes, for example: https://github.com/localsend/localsend/issues/907#issuecomment-1848828103

So the final filename would look something like this: LocalSend-1.14.1--google-play-signature.apk

And in the release notes description, specify that this is the Google Play version, as there is currently an issue with Google "rejecting updates". Additionally, it will serve for faster/manual updates. You could also create a topic similar to the one you made to inform about the issue related to app placement in the [Microsoft Store](https://github.com/localsend/localsend/issues/220), to inform users and say that for now, they will have to update the app manually. Because uninstalling/installing and using the app with several different signatures could eventually lead to problems, similar to what I had before https://github.com/localsend/localsend/issues/321, https://github.com/localsend/localsend/issues/829

As for the version that you have been signing and posting to GitHub so far, I think it should also be kept because it is useful for many who do not need updates and are satisfied with the current state of the app. I myself use some apps in this way.

PR #1497, opened 2024-07-13
i18n: Minor fix for Ukrainian, Russian and English languages

sergd88:
In my opinion, this option will be more literary and correct, because the current option sounds somewhat technical.


Tienisto:
Hi, I've written it this way so that it stays in the same line (at least in English).

Old:
Simulator Screenshot - iPhone 15 Pro Max - 2024-07-13 at 16 14 14

New:
Simulator Screenshot - iPhone 15 Pro Max - 2024-07-13 at 16 13 59

Tienisto:
I will merge this anyways since it is hidden as advanced setting

sergd88:
> Hi, I've written it this way so that it stays in the same line (at least in English).

Yes, I've noticed that depending on the display width/resolution, the parameter description will take up 2-3 lines instead of 1 as before. However, I still believe this option is better, especially for beginners who aren't yet familiar with the application's capabilities and need a clearer explanation of the parameter.

PR #1517, opened 2024-07-16
Fix the existing translation and add missing translation for Thai language

toopkungza:
Great app by the way! 👍


Tienisto:
Thanks!

PR #1521, opened 2024-07-16
i18n: Small fix for Ukrainian, Russian and English languages

sergd88:


sergd88:
It seems I messed up, and this fix also includes the previous one. I apologize for that.

Tienisto:
Thanks, I think the easiest solution is to delete your fork and create a new one

PR #1524, opened 2024-07-17
i18n: Final, minor fixes for Ukrainian, Russian, and English languages

sergd88:
I think these are the final corrections regarding translations until the next update.


Tienisto:
No worries, I don't care if more fixes come in since they improve the quality of the app :)

PR #1574, opened 2024-07-28
i18n: Translation corrections for Ukrainian, Russian, and English languages

sergd88:
Please, could you clarify in which cases I might see line 32: "reset": "Reset"? Depending on the context, it could be translated as a noun or a verb. In Ukrainian and Russian, this results in different endings. Therefore, I'd like to see where this parameter appears in order to translate it correctly.

This correction includes many changes because I finally took the time to review the entire translation for Ukrainian and Russian line by line. I hope this minimizes further minor corrections.

Also, in my opinion, instead of line 155: "Make sure that all devices are on the same Wi-Fi network and share the same configuration...", it would be better to use "Please make sure that all devices are on the same Wi-Fi network and share the same configuration...". The prefix "Please" adds more politeness to the request. That's why I adjusted this line.


sergd88:
Also, regarding block 338 of the line. The "title" is indicated as "Enter address," which is fine, but as a tooltip, it looks strange, since this is essentially a method of transmission, as far as I understand — manual. In my opinion, it would be worth adding a separate line, such as "description," with a label like "Manual sending," which will be displayed when hovering over it from the "Send" tab. And when clicking on it and opening the actual sending window, the current title "Enter address" is quite appropriate, as it is exactly what needs to be filled in.

Additionally, it would be worth adding a "description" for the "Refresh" button, as it is the only one without a tooltip, which also looks somewhat odd.

Tienisto:
Bildschirmfoto 2024-07-28 um 18 37 15

sergd88:
Oh, thank you very much. I was translating in the wrong context. I'll also make changes to the English language, because in my opinion, this button is better named "Undo changes", following the analogy with Microsoft Office, Git, etc.

sergd88:
Oh, thank you very much. I was translating in the wrong context. I also made changes to the English language, because in my opinion, this button is better named "Undo changes", following the analogy with Microsoft Office, Git, etc.

sergd88:
Oh, thank you very much. I was translating in the wrong context. I also made changes to the English language, because in my opinion, this button is better named "Undo changes", following the analogy with Microsoft Office, Git, etc.

sergd88:
Oh, thank you very much. I was translating in the wrong context. I also made changes to the English language, because in my opinion, this button is better named "Undo changes", following the analogy with Microsoft Office, Git, etc.

Tienisto:
Thanks!

PR #1592, opened 2024-07-30
Update _missing_translations_fr.json

Nixuge:
Here we go again.
Meaning changes:
- solution: changed meaning a bit, you can google translate for literal translation changes
- packagers: "emballeur" has a more irl vibe (eg packing w cardboard), "empaqueteurs" is ok-ish, maybe a bit too french but keeping "packagers" feels too english so would say that's the best choice


Tienisto:
Thanks!

PR #1616, opened 2024-08-08
Removed deprecated code from Theme.dart

hozefa-hs:
This PR removes deprecated code from Theme.dart.
Replaces MaterialStateProperty with WidgetStateProperty.


Tienisto:
LocalSend currently uses an older verion of Flutter.

PR #1622, opened 2024-08-10
Added Filipino Translation

leanwell:
Added Filipino Translation in README.md and README_TL.md
Added Translator's name to `AppLocale.filPH`
I used the ISO 639-1 tag 'TL'. Please correct me if this is the wrong tag.


Tienisto:
Since `PH` is used in the translation file, I think that `PH` would be better than `TL`

leanwell:
> Since `PH` is used in the translation file, I think that `PH` would be better than `TL`

I edited it, ty for corrections.

PR #1650, opened 2024-08-18
i18n: Update for missing translations/small correction of missing translations (Exit instead of Exit LocalSend) for all languages

sergd88:


sergd88:
I also replaced line 380 for German from "LocalSend beenden" to "Beenden" in line with other translations. If this isn't suitable for you, please, make the correction. Sorry for the inconvenience.

PR #1651, opened 2024-08-18
Update Polish translation

Matthaiks:


sergd88:
@Matthaiks Sorry for the inconvenience. For Windows, it's recommended to use just "Exit" without the app name (if you've Windows, you can verify this by checking other apps in the system tray, especially those from Microsoft). I've already sent a corresponding pull request (which hasn't been approved yet), but you added the translation faster. If it's not too much trouble, it would be great if you could make the necessary correction (380 line: "closeWindows"). Thank you in advance.

Best regards,
Sergiy Tkachuk

Issue #1625, opened 2024-08-10
Feature request: The parameter name "Enter Address" sounds somewhat out of place


sergd88:
### Is your feature request related to a problem? Please describe.

The parameter name "Enter Address" sounds somewhat out of place, since this is essentially a method of transmission, as far as I understand — manual. In my view, it would be worth adding a separate line, such as "description", with a label like "Manual sending", which would appear when hovering over it from the "Send" tab.

![1](https://github.com/user-attachments/assets/5cb70193-12ef-49d0-b734-edf2eba418de)

When clicking on it and opening the actual sending window, the current title "Enter Address" is quite appropriate, as we need to fill in the address/hashtag field.

![2](https://github.com/user-attachments/assets/a6c2de50-7b26-4c88-8344-46a88ee86260)

### Describe the solution you'd like

Described above.

Tienisto:
Thanks. I've added a separate translation for the button

PR #1653, opened 2024-08-18
i18n: Correction for English language

sergd88:
As for me "Search devices" sounds more casual.


sergd88:
By the way, just a reminder, if you've the time, please add an additional line of description for tooltip [1625](https://github.com/localsend/localsend/issues/1625). Judging by the fact that different acquaintances have asked me several times, it seems that newcomers aren't entirely clear about the purpose of the "Enter Address" mode.

PR #1698, opened 2024-08-24
Implemented feature to accept quickSave if sender is in favorites

Davte:
Implemented feature discussed in #861 and similar: when this option is set in settings, accept automatically requests coming from devices added to favorites.


Tienisto:
Similar to #1305
I don't know about this... it might be useful for most use cases but it is not completely secure

harveybolton:
> Similar to #1305 I don't know about this... it might be useful for most use cases but it is not completely secure

It's more secure than enabling "Quick Save" to auto-accept any inbound files from other devices on the network, and that's good enough for me. This would be a very welcome feature.

Tienisto:
Yeah, you are right. I guess a dialog mentioning this should be enough for the users.

Issue #1615, opened 2024-08-08
FR: Better macOS interoperability - drag and drop files to the menubar icon


ShlomoCode:
### Is your feature request related to a problem? Please describe.

The **ideal** goal for LocalSend is to be as simple to use as AidrDrop.
Currently the fastest way to send files with LocalSend is to leave the app open in the Dock and then drag the files to the Dock icon.

However, many users prefer a cleaner Dock, especially for apps used intermittently. Constantly visible apps can create visual clutter and potentially distract from primary tasks. This preference aligns with the popularity of menu bar apps in the macOS ecosystem, especially for quick-access\utility tools.

### Describe the solution you'd like

It would be very convenient if it was possible to leave the application running in the background (there is already such an option in the LocalSend settings), and when you want to send a file - just drop it on the menu bar icon.

### Describe alternatives you've considered

https://github.com/localsend/localsend/issues/586 pretty much serves this purpose, but I think throwing to the menu bar icon is more convenient as it's in one motion - just grab the file and drop it.

### Additional context

A macOS app I know of that has this feature is CleanShot X - any image can be dragged to the menu bar icon, and it will open in CleanShot's built-in editor.

ShlomoCode:
@Tienisto Can I suggest swift code for this?

I don't know how to integrate it with Flutter, but it works flawlessly as a native macOS app:

```
import SwiftUI
import Cocoa

@main
struct TestDragDropMenuBarApp: App {
    @NSApplicationDelegateAdaptor(AppDelegate.self) var appDelegate
    
    var body: some Scene {
        Settings {
            EmptyView()
        }
    }
}

class AppDelegate: NSObject, NSApplicationDelegate {
    var statusItem: NSStatusItem?
    
    func applicationDidFinishLaunching(_ notification: Notification) {
        statusItem = NSStatusBar.system.statusItem(withLength: NSStatusItem.squareLength)
        
        if let button = statusItem?.button {
            button.image = NSImage(systemSymbolName: "globe", accessibilityDescription: "Hello World")
            
            let dragView = DraggableView(frame: button.bounds)
            button.addSubview(dragView)
            
            dragView.translatesAutoresizingMaskIntoConstraints = false
            NSLayoutConstraint.activate([
                dragView.topAnchor.constraint(equalTo: button.topAnchor),
                dragView.leadingAnchor.constraint(equalTo: button.leadingAnchor),
                dragView.trailingAnchor.constraint(equalTo: button.trailingAnchor),
[...]

PR #1716, opened 2024-09-01
feat(macos): add drag-and-drop support for menubar icon

ShlomoCode:
closes https://github.com/localsend/localsend/issues/1615

Supports dragging files only (not strings), as adding support for strings requires changes on the flutter side that I don't have enough knowledge about.


Tienisto:
Thank you! This works as expected!

PR #1730, opened 2024-09-04
fix: selected files box padding in RTL lang

ShlomoCode:
before:
CleanShot 2024-09-04 at 16 00 20@2x
after:
CleanShot 2024-09-04 at 15 57 49@2x


Tienisto:
Thanks!

PR #1731, opened 2024-09-04
fix: file picker options padding in RTL lang

ShlomoCode:
before:
CleanShot 2024-09-04 at 16 50 50@2x
after:
CleanShot 2024-09-04 at 16 46 03@2x


Tienisto:
Thanks!

PR #1732, opened 2024-09-04
refactor(macos): replace deprecated `@NSApplicationMain` with `@main`

ShlomoCode:


Tienisto:
Thanks for the change. However, I get a different `LastUpgradeVersion` and `LastUpgradeCheck` (that is the value currently in the repo). I think it might be related that you are using a newer Flutter version?
The Flutter version is currently fixed to 3.13 to support Windows 7 (there are plans to drop this support soon)

ShlomoCode:
`LastUpgradeVersion` It's not the Flutter version, it's the newest Xcode version the project was opened with

Tienisto:
Okay, I found the reason. When starting the app from XCode, it is fine (I have XCode 15.4), but if I start from command line with `fvm flutter run`, then a script provided by Flutter resets the version to `1430` again. It seems that Flutter updates this version for compatibility reasons.

Issue #586, opened 2023-07-23
Feature request: Sharing plugin for macos


ShlomoCode:
### Is your feature request related to a problem? Please describe.

To send files, open the software, and drag them into the software window.
It's not bad, but it would be much nicer if it was possible to right-click > localsend sharing, just like Apple's AirDrop.

### Describe the solution you'd like

The Gifski app (https://github.com/sindresorhus/Gifski) contains a sharing plugin - you can right-click File > Sharing > Gifski, and thus send a file to the software.
This thing will be very useful for localsend, to be able to share file(s) with the software, similar to sharing with the built-in AirDrop.
A short video for demonstration:

https://github.com/localsend/localsend/assets/78599753/91922928-91d0-4f41-9b58-a9487d6ac615

ShlomoCode:
It would also be useful to allow dragging a file to the LocalSend icon in the menu bar
Similar to CleanShot app

NayamAmarshe:
Yes! This would be great!

RunAway189:
Yes, please. The context menu option would be very useful

image72:
@Tienisto have any plan support this feature?

Tienisto:
> @Tienisto have any plan support this feature?

Yes, but unfortunately, there is no ETA

Bob-SF:
Would love access to LocalSend via the Mac share menu.

ShlomoCode:
Ref: https://github.com/sindresorhus/Gifski/pull/156

PR #1739, opened 2024-09-06
feat(macos): add share extension

ShlomoCode:
Added LocalSend to [macOS Share menu](https://support.apple.com/en-jo/guide/mac-help/mh40614/mac) (currently for files only)

closes https://github.com/localsend/localsend/issues/586


ShlomoCode:
@Tienisto If you want to help me with the Dart side I can also add support for strings sharing. are you interested?

Tienisto:
Thank you!
It seems to properly select the file. However, I've noticed that the permission is not granted when triggered via "Share".
However, when I drag-and-drop the file into the app / app icon / menubar icon, then the permission is granted and I can send the same file via "Share".

ShlomoCode:
Thanks for catching this! I will investigate it

ShlomoCode:
@Tienisto I pushed a patch, now it should work properly. For technical details and traces of research, please see the comments I added.

Note this comment: [`de35bc8` (#1739)](https://github.com/localsend/localsend/pull/1739/commits/de35bc89a451d3cf5b8382b5bdc3621b2e94dde6#diff-1b28167c9d01216d1e06ec90fdc49e307367adb382d0b0530ad596efa2bc1ae6R28)
I will need you to tell me where the flutter side can send a message to the macos side that the handle to the file is no longer needed, so that we can call `stopAccessingSecurityScopedResource`.

Tienisto:
Thanks! It works! I will merge this and will add the call to stopAccessingSecurityScopedResource later

ShlomoCode:
Great! I want to add support for text sharing as well, can you please give me a small example snippet how do I add a text snippet to the queue from `macos_channel.dart`?

Tienisto:
Similarly to `getPendingFilesStream`, you would need to listen to another stream right at the app start.
For example, you can call `getMessageStream` right below `getPendingFilesStream` (https://github.com/localsend/localsend/blob/main/app/lib/config/init.dart#L198)

```dart
getMessageStream().listen((message) {
        ref.redux(selectedSendingFilesProvider).dispatch(AddMessageAction(message: message));
        ref.redux(homePageControllerProvider).dispatch(ChangeTabAction(HomeTab.send));
      });
```

NayamAmarshe:
Amazing! Thank you for the PR!

PR #1750, opened 2024-09-08
feat(macos): add support for text sharing and dropping

ShlomoCode:
Added support for sharing web URLs and texts using macOS Share menu, and added support for drag and drop text onto the menu bar item and the Dock icon

Completion of #1716 and #1739


Tienisto:
Thank you!

PR #1762, opened 2024-09-09
Added Slovak translation

dodog:
Added Slovak translation file


dodog:
It is a great app, thank you.

PR #1766, opened 2024-09-09

sergd88:
Since there is no contact list but rather a favorites list, I think it should specify "favorites list" to avoid confusing users (changes for the English).


Tienisto:
Thank you!

PR #1768, opened 2024-09-10
Update strings_zh-CN.json

nkh0472:
Want to know how to translate:
```
  "settingsTab": {
    "receive": {
      "quickSaveFromFavorites": "@:general.quickSaveFromFavorites"
    }
  },
  "dialogs": {
    "quickSaveFromFavoritesNotice": {
      "title": "@:general.quickSaveFromFavorites"
  }
```

---

After searching, there are already some strings strat with "@" and not translated. I'll keep them untouched.


Tienisto:
Thanks. String keys starting with `@:` are linking to an existing translation. Usually, they don't need to be changed

Issue #1688, opened 2024-08-23
FR: macOS - Show Transfer Progress in Dock Icon


ShlomoCode:
### Is your feature request related to a problem? Please describe.

It would be helpful to show a progress line of the file transfer in the Dock icon in macOS, so that you can switch to another window and see the transfer status at a glance

### Describe the solution you'd like

Show file transfer progress in the Dock icon

### Describe alternatives you've considered

display a percentage badge instead of a tick around the icon

### Additional context

A popular Swift library for this: https://github.com/sindresorhus/DockProgress

Example ([Gifski](https://github.com/sindresorhus/Gifski)):

https://github.com/user-attachments/assets/c141dbb5-c001-443e-be4e-939b72e50c00

ShlomoCode:
I see that the feature has already been implemented in #1222 and canceled in https://github.com/localsend/localsend/commit/0ced63b88103f9556f0db234bd6970918cd22985. @Tienisto, could you elaborate on this more?

Tienisto:
@ShlomoCode I couldn't compile it somehow: https://github.com/NightFeather0615/macos_dock_progress/issues/1

ShlomoCode:
OK, We can directly use the original swift package (https://github.com/sindresorhus/DockProgress) in the native macOS Runner. Are you interested in such a PR?

Tienisto:
Sure. You can add any function to https://github.com/localsend/localsend/blob/main/app/lib/util/native/macos_channel.dart that invokes a native macos function

PR #1770, opened 2024-09-10
feat(macos): show transfer progress in Dock icon

ShlomoCode:
closes https://github.com/localsend/localsend/issues/1688

cleanshot_2024-09-08_at_19 33 24_2x

Flutter's Swift Package Manager support is only from Flutter version 3.24, and we are stuck on 3.13, so we lose things like https://github.com/sindresorhus/DockProgress/issues/1 (DockProgress dropped CocoaPods support in v4.0.0)
Once we're free, switch to the latest version of DockProgress using the Swift package manager (same as the Default dependency).

### DIFF
--- a/app/lib/util/native/macos_channel.dart
+++ b/app/lib/util/native/macos_channel.dart
@@ -18,6 +18,14 @@ Future<void> setupStatusBar() async {
   });
 }
 
+Future<void> updateDockProgress(double progress) async {
+  if (defaultTargetPlatform != TargetPlatform.macOS) {
+    return;
+  }
+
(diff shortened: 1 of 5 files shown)


Tienisto:
Thanks for the PR! Even if we upgrade to 3.24, I don't know if Flutter's Swift Package Manager support is stable yet.

ShlomoCode:
https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-app-developers
Anyway, this PR still uses pods

ShlomoCode:
I solved the conflicts

Tienisto:
Thanks!

PR #1771, opened 2024-09-10
refactor: simplified `methodChannelInitialized` awaiting

ShlomoCode:
refactor of https://github.com/localsend/localsend/commit/cf48c539fdf36902329f450a8513ace25e9ec7f2

The main point is that the `Defaults.observe` callback is called not only for future changes, but is called first for the initial value.


Tienisto:
Alright!

Issue #1779, opened 2024-09-12
UI Enhancement: Improve display of advanced options on checkbox selection


AyanDas-99:
### Describe the bug

When selecting the "Advanced Settings" checkbox, the advanced options are currently inserted between the existing options, which disrupts the flow of the form, causing a confusing user experience.

You can see, the 'Send' setting section is also added, but the user may completely miss it.

### To reproduce

1. Go to 'Settings'
2. Click on  'Advanced settings'

### Expected behavior

I suggest adding the advanced options directly below the checkbox. This way users will clearly understand that these options are related to the "Advanced Settings."

### Screenshots
![Screenshot from 2024-09-13 00-25-09](https://github.com/user-attachments/assets/ffbd05da-be8d-4755-b4fa-151c407f0ed9)
![Screenshot from 2024-09-13 00-25-32](https://github.com/user-attachments/assets/17c49bd7-4e0e-49df-bd40-e2f3d1e21f1f)

PR #1780, opened 2024-09-12
fix: move advanced setting entries below advanced setting checkbox

AyanDas-99:
fixes: #1779
Brought these setting entries under the Advanced settings checkbox
- 6 entries from the Network section
- The whole Send section

Screenshot from 2024-09-13 00-26-58


Tienisto:
In my opinion, "Other" should always be last. I know that the check box is not good UX but I dislike moving "Other" up.

AyanDas-99:
checking advanced options just populates the existing sections and is difficult to tell what are actually the advanced options. So, I thought separating the advanced options would be better.
What is your opinion @Tienisto ?

PR #1783, opened 2024-09-16
feat(macos): show checkmark/xmark in Dock icon on transfer success/failure

ShlomoCode:
closes https://github.com/localsend/localsend/issues/1772

Uses https://github.com/localsend/localsend/blob/main/app/macos/Runner/Assets.xcassets/AppIcon.appiconset/logo-1024-mac-256.png plus checkmark\xmark from Apple's SF Symbols, and adding A little transparency and shadow using Adobe Photoshop. Inspired by WebStorm and Slack (see Materials).

Demo:

https://github.com/user-attachments/assets/695f8675-9d3f-4977-ad10-54fc4867c8dc

Materials:
[logo.zip](https://github.com/user-attachments/files/17007881/logo.zip)

PR #1785, opened 2024-09-17
fix: dock progress shows only once

ShlomoCode:
Currently, changing the Dock icon (#1783) breaks DockProgress, so that after one completed transfer, the progress is no longer displayed in the Dock.
The behavior is the same in DockProgress v v4.3.0 but in v4.3.1 it has been fixed, so we have to use the Swift Package Manager through Xcode at the moment to resolve this package (because the cocoapods version will no longer be updated), and then it will be possible to build normally using the flutter command.

### DIFF
--- a/app/lib/util/native/taskbar_helper.dart
+++ b/app/lib/util/native/taskbar_helper.dart
@@ -26,7 +26,6 @@ class TaskbarHelper {
       if (_isWindows) {
         await WindowsTaskbar.setProgress(digestedProgress, digestedTotal);
       } else if (_isMacos) {
-        print(progress);
         await updateDockProgress(progress / total);
       }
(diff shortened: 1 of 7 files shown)


Tienisto:
Weird, I still get this bug

ShlomoCode:
Did you make sure you are using the SPM version and not the Pod version?
Try:
1. "Clean Build Folder" in Xcode and run
2. `fvm flutter clean && fvm flutter pub get && fvm flutter build macos`
3. "Resolve Package Versions" in Xcode
4. re-build

Tienisto:
nvm, `gh pr checkout` didn't work so I deleted the old branch having the same name. Now it seems to work. Thanks!

Issue #1242, opened 2024-03-29
Feature request: start LocalSend in menu bar MacOS


Tsjoep:
### Is your feature request related to a problem? Please describe.

On login, LocalSend starts as an open window in the dock.

### Describe the solution you'd like

On login, start LocalSend minimized in the menu bar instead as an open window in the dock.

### Additional context

For example, just as Google Drive starts minimized in the menu bar on login.

Mennaruuk:
This is a good feature request, but I hope it doesn't become the default behavior, since some LocalSend users will be confused why no window appears after launching the program and resort to another file-sharing app. Hopefully there is an option to enable/disable it as needed.

Tienisto:
Added in #1467

Tsjoep:
> Added in #1467

I don't see an option to enable this on the MacOS app? Am I missing something?

Tienisto:
@Tsjoep sorry, this only works in the DMG version or homebrew. If you installed LocalSend via App Store, then it currently does not work

Tsjoep:
**Feature request update**

As of now, LocalSend can start minimized (hidden) in the dock - but not in the menu bar.
It can however be minimized to the menu bar when closing.

Would it be possible to add an option to autostart LocalSend hidden in the menu bar instead of the dock?

PR #1793, opened 2024-09-18
feat(macos): add sandboxed launch at login for App Store version

ShlomoCode:
https://github.com/localsend/localsend/issues/1242

Not used CocoaPods since the owner [says](https://github.com/sindresorhus/LaunchAtLogin-Legacy/releases/tag/v3.0.0#:~:text=It%20didn%27t%20work%20correctly%20anyway.) taht "It didn't work correctly".

ref https://github.com/sindresorhus/LaunchAtLogin-Legacy/issues/33#issuecomment-2253207190, ref https://github.com/hisaac/CenterMouse/blob/main/CenterMouse/Sources/AppDelegate.swift

### DIFF
--- a/app/lib/config/init.dart
+++ b/app/lib/config/init.dart
@@ -105,7 +105,11 @@ Future<RefenaContainer> preInit(List<String> args) async {
     if (args.contains(startHiddenFlag)) {
       // keep this app hidden
       startHidden = true;
+    } else if (defaultTargetPlatform == TargetPlatform.macOS) {
+      startHidden = await isLaunchedAsLoginItem() && await getLaunchAtLoginMinimized();
+    }
 
+    if (startHidden) {
       unawaited(hideToTray());
     } else {
       await WindowManager.instance.show();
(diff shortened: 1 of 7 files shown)


Tienisto:
I am thinking to upgrade to the latest Flutter version so that we can use the Swift Package Manager.
I think this would be easier to maintain your recent macos changes

ShlomoCode:
Sounds good!
All that's left here to make it ready to merge is to keep the AutoStartHidden preference for macOS as now it's not determined by the file contents. Is it worth doing it in Flutter or in the macOS Runner? Thoughts?

Tienisto:
I think you can do it in `UserDefaults`. There is not much of a difference though so pick the solution you are comfortable with!

ShlomoCode:
I think it's ready

Tienisto:
Nice, thank you!

ShlomoCode:
@Tienisto I think we need to add deletion of the old agent to prevent duplication for those who already activated the old way?

PR #1797, opened 2024-09-22
feat(macos): open the main window when the app is reopened

ShlomoCode:
When the app is minimized to the menu bar, the expected behavior when reopening the app from the Launchpad is that the app window will open.
It's also useful for users who accidentally activated "Minimize to the System Tray/Menu Bar when closing" and now don't understand how to get the app.


Tienisto:
Nice!

Nicediver-bit:
Ok

SahilBhalani:
That's a very nice update @ShlomoCode

Nicediver-bit (commented):
Very good

PR #1813, opened 2024-09-25
refactor: remove platform checks in macos_channel.dart

ShlomoCode:
thoughts?

### DIFF
--- a/app/lib/util/native/macos_channel.dart
+++ b/app/lib/util/native/macos_channel.dart
@@ -1,6 +1,5 @@
 import 'dart:async';
 
-import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
 import 'package:localsend_app/gen/strings.g.dart';
 import 'package:localsend_app/util/native/taskbar_helper.dart';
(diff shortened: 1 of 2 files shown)


Tienisto:
Hi. There are native `assert`s in Dart that might be better. So it doesn't crash in production and only throws during debug mode

ShlomoCode:
But if we call function in wrong case, we'd prefer a crash so we know there's a problem here, over silencing the error, right?

ShlomoCode:
What is meant by abstraction layer?

Tienisto:
Looking at the code again, I think we can remove the checks altogether since it is clear that any function in `macos_channel.dart` should be called on macOS environments only. The function will throw an error regardless of your added exception.

There are already `taskbar_helper.dart` and `autostart_helper.dart` that are platform independent.
However, `pendingFilesStream` and `pendingStringsStream` should be extracted to an extra file also (for example `file_listener.dart` where calls are proxied to `macos_channel.dart` for macOS and are no-op for any other platform.

ShlomoCode:
@Tienisto See now

PR #1816, opened 2024-09-25
i18n: Updated the Italian translation for consistency.

LandOfDar:
Improved language consistency to better adapt to the different targeted ecosystems. 
Fixed some capitalization mistakes for iOS and Wi-Fi, and characters between parentheses/after the “:” sign. 
After reviewing it for several times today, I think it's a good improvement based on the original translation, which was already pretty clear.


Tienisto:
Thank you

Issue #977, opened 2023-12-03
Ask user to insert a device name on first launch instead of generating a random one


mijorus:
### Is your feature request related to a problem? Please describe.

No

### Describe the solution you'd like

A pop-up welcome screen asking the user to provide a name for the current device

sergd88:
I disagree that randomly-generated names are quite ugly. Rather, they just appear somewhat informal and less professional. However, @Tienisto mentioned somewhere that in this regard, he's somewhat inspired by the approach of [Snapdrop](https://snapdrop.net/). [ShareDrop](https://www.sharedrop.io/), for example, uses animal icons. In general, in such cases, I recall Ubuntu, which, despite everything, remains the most popular Linux distribution and is used by many business clients. Their approach to release names, such as **Disco Dingo** 😂, reminds that it isn't necessary to always be so formal 😉

As for a pop-up welcome screen asking user to provide a name for the current device, it's certainly a great idea, though not a new one. The key isn't to greet the user immediately with a name input field 😄

mijorus:
> I disagree that randomly-generated names are quite ugly. Rather, they just appear somewhat informal and less professional. However, @Tienisto mentioned somewhere that in this regard, he's somewhat inspired by the approach of [Snapdrop](https://snapdrop.net/). [ShareDrop](https://www.sharedrop.io/), for example, uses animal icons. In general, in such cases, I recall Ubuntu, which, despite everything, remains the most popular Linux distribution and is used by many business clients. Their approach to release names, such as **Disco Dingo** 😂, reminds that it isn't necessary to always be so formal 😉
> 
> As for a pop-up welcome screen asking user to provide a name for the current device, it's certainly a great idea, though not a new one. The key isn't to greet the user immediately with a name input field 😄

With "ugly", I meant unprofessional.

Ubuntu is not a tool that an average user might use, but rather it's designed for geeks who are used to these kinds of jokes.

While I agree that a pop-up window might not be the best UX, I would at least try to generate a good-enough name based on real device information.

On android, we could use the device name (the same that appears under Bluetooth settings for example)

sergd88:
> > I disagree that randomly-generated names are quite ugly. Rather, they just appear somewhat informal and less professional. However, @Tienisto mentioned somewhere that in this regard, he's somewhat inspired by the approach of [Snapdrop](https://snapdrop.net/). [ShareDrop](https://www.sharedrop.io/
> 
[...]

PR #1829, opened 2024-09-27
UX Feature: Add alias re-generation and system name buttons to settings page

Nolle10:
This PR addresses some ideas from issue #977 by enhancing the user experience in the network settings tab. It introduces two new buttons to streamline alias management:

**System Name as Alias:** Allows users to quickly set their system's local hostname as the alias for easier device identification.
**Random Alias Generator:** Provides an option to generate a new random alias for users who prefer another identifier.

**Preview of Functionality**
Image of two buttons
Tooltip for Generating Alias
Tooltip for Using System Name

**Key Changes**
Added two IconButtons with tooltips for generating random aliases and using the system's name.
Automatically updates the alias text field and persists the change using settingsProvider.

**Tested on:**
- Windows 10
- Iphone 15


Tienisto:
I think replacing the textfield with a dialog would be better because the 2 added buttons make the page less consistent

Nolle10:
I have changed the textfield to a dialog box, to ensure consistency.

image

image

Tienisto:
Thank you!

PR #1837, opened 2024-09-28
Add persistence for 'Advanced Settings'

Nolle10:
Persist the boolean for 'Advanced Settings', making it more user-friendly.
The thought behind the PR is that users enabling 'Advanced Settings' would like to always have it enabled, and not have to always enable it when closing and opening the application.

**Key Changes:**
- Add persistence for 'Advanced Settings' boolean

**Tested on:**
Windows 10


app/lib/provider/persistence_provider.dart
```diff
@@ -161,6 +162,10 @@ class PersistenceService {
       await prefs.setString(_securityContext, jsonEncode(generateSecurityContext()));
     }
 
+    if (prefs.getBool(_advancedSettingsKey) == null) {
+      await prefs.setBool(_advancedSettingsKey, false);
```
Tienisto, line 166: No need to initialize if there is a default value

Tienisto:
Thank you

Issue #1866, opened 2024-10-05
Add clipboard as a picker option in iOS


AnessZurba:
### Is your feature request related to a problem? Please describe.

Right now, the ios version doesn't have "clipboard" in the send screen

### Describe the solution you'd like

Add clipboard to the list

### Additional context

This seems to be a one-liner, but I have to setup the project on macOS to test it. Is there a reason clipboard was removed from iOS?

AnessZurba:
Ok, even after setting up on macOS. It seems I can't build the ipa because I'm not part of the team - provisioning profile stuff

AnessZurba:
I managed to run the app
I opened a PR https://github.com/localsend/localsend/pull/1873

PR #1754, opened 2024-09-08
feat: support clipboard for android

Seidko:
### description
Support clipboard for android, add paste button.
close #1753 

This pr need wait upstream to update. It modify lot of dependencies and need more test.

PR #1873, opened 2024-10-06
file_picker: add clipboard option to iOS.

AnessZurba:
Following https://github.com/localsend/localsend/issues/1866
I did manage to build the app to test the change. It seems to be working when running on ipad. It also seems to be working on iphone simulator. 
Please let me know if there is any reason not to do this change, as it seems like it was left out on purpose.

Simulator Screenshot - iPhone 16 - 2024-10-06 at 22 28 38

### DIFF
--- a/app/lib/util/native/file_picker.dart
+++ b/app/lib/util/native/file_picker.dart
@@ -70,6 +70,7 @@ enum FilePickerOption {
       return [
         FilePickerOption.media,
         FilePickerOption.text,
+        FilePickerOption.clipboard,
         FilePickerOption.file,
         FilePickerOption.folder,
       ];


Tienisto:
Thanks! I think the reason was that "Clipboard" has been tested on Desktop only (#674). Nobody added this on mobile platforms because there was no demand (until now). For Android, there was actually a change needed (#1754).

AnessZurba:
@Tienisto Thanks! 
Do you know when would this be reflected in the App Store version? What is your release cadence for App Store?

AnessZurba:
I see, sounds exciting
I'm also very excited regarding the cli project you are working on it. I hope this becomes the de facto standard at some point :)

Issue #1608, opened 2024-08-06
Feature request - Always show transfer speed by default


bibsp:
### Is your feature request related to a problem? Please describe.

Transfer speed is now hidden. Make it visible always by default or give an option to enable it.

### Describe the solution you'd like

Transfer speed is now hidden. Make it visible always by default or give an option to enable it.

PR #1920, opened 2024-10-21
Show Transfer Speed (Feature Request: #1608)

KISHORE-KUMAR-S:
Following Feature Request: #1608 

The transfer speed is now visible by default while sending or receiving the file irrespective of languages.

WhatsApp Image 2024-10-21 at 15 16 43
WhatsApp Image 2024-10-21 at 15 12 24
WhatsApp Image 2024-10-21 at 15 12 15

### DIFF
--- a/app/devtools_options.yaml
+++ b/app/devtools_options.yaml
@@ -1,3 +1,0 @@
-description: This file stores settings for Dart & Flutter DevTools.
-documentation: https://docs.flutter.dev/tools/devtools/extensions#configure-extension-enablement-states
-extensions:
--- a/app/devtools_options.yaml
+++ b/app/devtools_options.yaml
@@ -1,0 +1,3 @@
+description: This file stores settings for Dart & Flutter DevTools.
+documentation: https://docs.flutter.dev/tools/devtools/extensions#configure-extension-enablement-states
+extensions:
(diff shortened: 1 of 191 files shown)


Tienisto:
Besides the massive PR size, I don't think this should be permanent visible.

KISHORE-KUMAR-S:
> Besides the massive PR size, I don't think this should be permanent visible.

I can't have a CI/CD clearance for format. Is it due to massive pull request?

KISHORE-KUMAR-S:
> Besides the massive PR size, I don't think this should be permanent visible.

I would like to confirm that this should be optional?

KISHORE-KUMAR-S:
> Besides the massive PR size, I don't think this should be permanent visible.

Besides running "dart format ." command, I've formated all the files. This resulted a massive PR size.

KISHORE-KUMAR-S:
> Besides the massive PR size, I don't think this should be permanent visible.

I've formatted all the files, but the PR size is still pretty big after running 'dart format .'. It looks like the CI/CD format check isn't working. Fast transfer speeds are really important for a good user experience and real-time collaboration.

Tienisto:
The speed is important to "nerds" like us. Most users only need to know how long to wait. I also dislike the fact that there are 2 information in brackets now and the higher probability that we need to use 2 lines instead of 1 - both are making the UI worse.

PR #1924, opened 2024-10-22
Show Quick Save and Quick Save Favorites Dialog once

KISHORE-KUMAR-S:
The Quick Save and Quick Save Favorites dialogs will now be displayed only once when the app is installed and the Quick Save and Quick Save Favorites settings are enabled for the first time in the Settings tab.


KISHORE-KUMAR-S:
I've used SharedPreferences for this.

PR #1940, opened 2024-10-26
Update strings_cs.json

RobertBobby007:


Tienisto:
This is not valid JSON. The braces `{` are there for a reason.

Issue #1942, opened 2024-10-26
"LocalSend.app would like to access data from other apps" popup on macOS 15


ShlomoCode:
![CleanShot 2024-10-26 at 20 55 38@2x](https://github.com/user-attachments/assets/095778ed-ed8f-446e-a787-313ab11702f1)

This popup appears every time the app is opened, because the permission is [per process](https://forums.developer.apple.com/forums/thread/742147?answerId=773022022#773022022).

I am still researching this, but here are some relevant links:

https://forums.developer.apple.com/forums/thread/758375
https://developer.apple.com/forums/thread/721701
https://developer.apple.com/documentation/macos-release-notes/macos-15-release-notes#System-Integrity-Protection

ShlomoCode:
OK, it seems that adding the team identifier as the shared group prefix solves it. I will open a PR

PR #1944, opened 2024-10-26
fix: "would like to access data from other apps" warning popup on macOS Sequoia

ShlomoCode:
closes https://github.com/localsend/localsend/issues/1942


Tienisto:
Thanks

PR #1961, opened 2024-10-28

sergd88:
I've a question: is there any character limit? I think it would be better to use more detailed explanations. Therefore, I made some changes to the English translation.

If for any reason this isn't acceptable to you, please reject this pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 [Homepage][homepage] • [Discord][discord] • [GitHub][github] • [Codeberg][codeberg]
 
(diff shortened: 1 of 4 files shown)


Tienisto:
Thanks! No there is no character limit.

For a bit more context: Here is the new UI of the "Receive" tab. When "Favorites" is selected, the dialog with the warning is shown. So the last part `by enabling the "Automatically accept requests" option.` is a bit confusing here. Also: Maybe "restriction" is a better (simpler) word than "hindrance".

Bildschirmfoto 2024-10-28 um 18 43 52

PR #2022, opened 2024-11-07
fix: keep screen awake during large file transfers

### LINKED ISSUE #1510: Please add feature "keep the screen on" to iOS app

When sending large files, iOS will turn the screen off after certain amount of idle time and kill the networking activity.

Simply let the user opt into (or opt out of) "keep the screen on" in the setting panel.

_No response_

### Additional context

_No response_

anishkargaonkar:
Closes #1510 where the screen would turn off while transferring large files, disrupting the transfer process. 

To solve this, `WakelockPlus.enable()` is called periodically to ensure the device remains awake throughout the file transfer session and to handle cases where the OS may release the wakelock unexpectedly. According to the WakelockPlus's documentation, there is no negative impact in calling `WakelockPlus.enable()` more often.


Tienisto:
Nice!

Issue #2043, opened 2024-11-14
Bug report - share with web link, if user replace a new file with a same file name; then download file will not success


MisterChangRay:
### Describe the bug

on windows client, share with web link, if user replace a new file with a same file name; then download file will not success

### To reproduce

way1:
1. share a file with web link
2. use chrome download file and will be succes
3. close share with link and reupload a new file with same name
4. reopen share with web link
5. use chorme download file and not success

way2:
1. share a file with web link
2. use chrome download file and will be succes
3.  open file directory and use another file repleacement, use a same file name
5. refresh chrome and click download file 
6. download not be success

### Expected behavior

download success

PR #2044, opened 2024-11-14
fixbug #2043

MisterChangRay:
bugfix for #2043 

because the file length  will cached on file open; 
so http reponse length use older data(if file be replace with same name on download);
so the chrome download will be fail;

fix:
- on download event reload file length


Tienisto:
LGTM

Issue #2078, opened 2024-11-22
Cannot open file only when sending images from Windows to Android


RealNath:
### Describe the bug

Whenever I want to send images from Windows to Android, the image won't open. It keeps throwing 'Could not open "[file location]". Has this file been moved, renamed or deleted?' . But other file types from Windows to Android (and vice versa) would work perfectly, and sending images from Android to Windows works too.

### To reproduce

1. Send image(s) from the Windows app (whether it's from manual, drag and drop, or the context menu's "Send to")
2. Select the Android device
3. On Android, tap "Open" or "Show in folder"
4. The error 'Could not open "[file location]". Has this file been moved, renamed or deleted?' occurs.

### Expected behavior

The image files should be openable or shown in folder.

### Screenshots

![Screenshot_2024-11-23-00-37-39-126_org localsend localsend_app](https://github.com/user-attachments/assets/6b2e1036-fbf3-4cf8-85be-1b990548d5d1)

### Desktop (please complete the following information)

- OS: Windows 11
- Version: 24H2 (Release Preview channel)

### Smartphone (please complete the following information)

- Device: Xiaomi 12T
- OS: Android
- Version: 14

Tienisto:
Yeah, it should open the gallery because "Save media to photos" is likely enabled.

uclk:
Same here. Do you actually have the image file in the destination folder? I checked, and there’s no such file there.

saurabh12453:
Yes i have checked will revert it soon cant access due to lack of network
issue will do by tomorrow..

PR #2088, opened 2024-11-25
Fix image file opening issue from Windows to Android

theguy000:
Related to #2078

Modify the `openUri` function in `FileOpener.kt` to handle image files in the cache directory.

* Add a check to move image files to a user-accessible directory before opening.
* Add a new function `moveFileToUserAccessibleDirectory` to handle the file move operation.
* Update the `openUri` function to use the new function for image files.


Tienisto:
Hi, this shouldn't be possible because the cached image is deleted right away when "Save media to gallery" is enabled.

PR #2092, opened 2024-11-26
feat: Convert Clipboard image to png

### LINKED ISSUE #872: Convert clipboard image from BMP to PNG

The current version sends screenshots from the clipboard in BMP format, but some software doesn't support BMP format.

Send the screenshot from the clipboard in PNG format.

_No response_

### Additional context

_No response_

BrianMwit:
fixes #872

Alternatively, we can request pasteboard to make this change. Their Mac [mac](https://github.com/MixinNetwork/flutter-plugins/blob/b2beb15590f7b6c38e650ae501923b9bd3e7849c/packages/pasteboard/macos/Classes/PasteboardPlugin.swift#L42-L47) and [Linux](https://github.com/MixinNetwork/flutter-plugins/blob/b2beb15590f7b6c38e650ae501923b9bd3e7849c/packages/pasteboard/linux/pasteboard_plugin.cc#L92) native codes are already converting image buffer to png.

pasteboard's bmp seems to discard alpha channel, so transparency is already lost.
Even with this limitation, I would guess this already covers most use cases

Screenshot:
image

I have only tested this change on Windows, but don't expect this to be an issue with any other platform

Apologies for the half-baked PR. I have never used flutter/dart before and thus don't know best practices.
Feel free to clean up this patch as needed.

### DIFF
--- a/app/lib/util/native/file_picker.dart
+++ b/app/lib/util/native/file_picker.dart
@@ -6,6 +6,7 @@ import 'package:file_selector/file_selector.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
+import 'package:image/image.dart' as img;
 import 'package:localsend_app/config/theme.dart';
 import 'package:localsend_app/gen/strings.g.dart';
 import 'package:localsend_app/model/cross_file.dart';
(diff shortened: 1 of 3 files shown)


Tienisto:
Nice!

Issue #2102, opened 2024-11-28
Make left click on tray icon open app, not context menu


n-bux:
Hi, love the app, thanks so much for it!

On Linux (Mint 20.3 Cinnamon) with LocalSend 1.16.1 (Flatpak install), 

after minimizing to tray,

when I click the the tray icon, I get the context menu, no matter if left or right mouse button.

As I see the code @ 
https://github.com/localsend/localsend/blob/0986cd105ef29fc5f86c777cac33c7de70d3bbb2/app/lib/widget/watcher/tray_watcher.dart#L36C1-L43C4 , it should only do that on macos, but it is doing it on (my) linux as well. 

Could this be fixed, so that a left click takes me straight to the app (`showFromTray();`)?

PR #2108, opened 2024-11-29
Fix tray icon click behavior on Linux

theguy000:
Related to #2102

Update `onTrayIconMouseDown` method to differentiate behavior for Linux and macOS.

* Call `showFromTray` for left-clicks on Linux and macOS.
* Display context menu only for right-clicks on Linux and macOS.


Tienisto:
On macOS, it should actually show the context menu on left clicks. With your change the behaviour on Windows platform is now incorrect. You did not fix the bug but introduced 2 new bugs on macOS and Windows.

Do you use AI? Because you should at least test your changes before submitting nonsense.

PR #2173, opened 2024-12-29
README translation in Ukrainian and Russian

sergd88:
I've made translations of the README for Ukrainian and Russian. I don't fully understand why, as in my opinion, it's not really necessary, especially for Ukrainian and Russian, and it makes the page a bit cluttered. But I saw that someone had already sent a pull request regarding the Russian translation, so it's better for me to translate it to avoid editing someone else's work later 😄

I also suggest arranging all available languages in alphabetical order, as I think it'll make it easier to navigate and find the one you need if it's available. However, English should still be placed first since it's the default language, and all new changes will primarily be reflected in it. Also, since there are already quite a few languages, it's worth labeling English as English (Default), which I've done, to emphasize that it's the default language, as I've encountered projects where English is available but it this isn't the main language.


Tienisto:
Thank you! Yes, the translated README files are currently unordered...

PR #2199, opened 2025-01-06
.sendignore basics done

chanmaoganda:
though most features could be achieved, some are wired

`**.dart` would not match /some.dart
`**/**.dart` would not match some.dart


chanmaoganda:
Currently only patterns like below is supported in dart glob package
image

we cannot use `/.git` or `.git/` to filter all contents in .git folder
Both `.git/*` and `.git/**/*` are required for filtering all content in .git folder.

chanmaoganda:
i wonder if we can pre-process patterns like `[.git/, /.git, .git/*, *.rs]`
for example, we can process `.git/*` to `[.git/*, .git/**/*]`
process `*.rs` to `[*.rs, **/*.rs]`

Tienisto:
Thanks. This should be ok for the first step!

PR #2204, opened 2025-01-09
feat: initial webrtc signaling prototype

### LINKED ISSUE #1703: Feature Request: Mesh Network and Device Discovery

Currently, devices must be in the same network in order to discover and transfer files to each other, which poses some limitations. Mesh network may help to solve this issue I think

Is it possible to support mesh network feature, which can allow devices to discover others across different networks and automatic routing to transfer files? This feature is just like what wireguard/tailscale does, but only need a simple implementation, not so complicated as they are.

This feature will be especially useful. For example, a computer with both wifi and LAN connections can allow other computers to transfer files to and from my phone.

Another example, maybe I can deploy headerless LocalSend on the server to support transfering files between off-site devices, as what it does in Airdrop.

Tienisto:
Closes #1703


itz-rj-here (approved):

Tienisto:
Only discovery working for now. I will merge this regardless to avoid future merge conflicts

PR #2247, opened 2025-01-31
chore: add gradient fadein effect

lingyicute:
I've added a gradient effect to the front page "alias" to make it consistent with the style of the other elements. I've tested it sufficiently to make sure there are no potential issues.


Tienisto:
This is intentional to show the important information (alias) first

PR #2304, opened 2025-02-22
Update zh-CN.json

EFLKumo:
I haven't reviewed the related changes for a while, but I found some issues:

1. The meaning of some words does not match the English version
2. Some expressions are very unnatural in Chinese, such as "需要一个 PIN", which might shows traces of machine translation 🤔
3. Some content may be confusing, I have made minimal and necessary explanations
4. "multicast" was translated as "多线程广播", which refers to "multi-threaded broadcasting" rather than "multicast".

This version fixes the above problems while ensuring no redundant modifications are made. Please check at your convenience :D
(btw, let me say a thank for all Chinese translation contributors and the maintainer for this pretty app : )

### DIFF
--- a/app/assets/i18n/zh-CN.json
+++ b/app/assets/i18n/zh-CN.json
@@ -46,7 +46,7 @@
     "infoBox": {
       "ip": "IP：",
       "port": "端口：",
-      "alias": "别名（设备名称）："
+      "alias": "设备名称："
     },
     "quickSave": {
       "off": "@:general.off",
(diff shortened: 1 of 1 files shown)


graphemecluster (approved):
LGTM, feel free to update zh-HK and zh-TW accordingly (or I’ll do it later)

EFLKumo:
> LGTM, feel free to update zh-HK and zh-TW accordingly (or I’ll do it later)

Thanks~
I'm not good at Traditional Chinese so you can take it from here.

PR #2356, opened 2025-03-16
Updated missing translations for fil-PH

BryanJames16:
Updated missing translations for fil-H localization.


BryanJames16:
Good Day @Tienisto !

Thank you for the merge! I would just like to ask if I still need to perform localization updates through `flutter pub run slang` before creating a PR? Or updating missing translations file would suffice? Thanks in advance!

Tienisto:
Only updating missing translations is enough

PR #669, opened 2023-09-06
fix: disable gtk client side decorations when the environment variable GTK_CSD exists and its value is 0

I-Want-ToBelieve:
I tested the following five situations, all as expected

```fish
GTK_CSD=0 flutter run
GTK_CSD=1 flutter run
GTK_CSD=test flutter run
unset GTK_CSD && flutter run
gtk3-nocsd flutter run
```

ref: 
gtk_csd: https://docs.gtk.org/gtk4/running.html#gtk_csd
gtk3-nocsd: https://github.com/PCMan/gtk3-nocsd

#666

PR #2360, opened 2025-03-16
Disable GTK3 header bar by default on Linux and use regular title bar instead

nixigaj:
I don't know why, but when Google created the Flutter Linux template, they decided to use the GTK3 header bar on Linux, even though almost no Flutter projects intend to place any buttons or functionality in the header bar. This results in the title bar being unnecessarily large in Wayland environments and GNOME in X11 mode.  

When a GTK window uses the header bar, it also does not respect the request from the Wayland compositor to draw server-side window decorations through the [XDG Decoration protocol](https://wayland.app/protocols/xdg-decoration-unstable-v1). This leads to non-native window decorations on KDE Wayland, Sway, and any other Wayland compositor that supports the protocol.  

This PR ensures that LocalSend always uses the regular GTK3 title bar unless the `GTK_CSD` environment variable  [introduced in this PR](https://github.com/localsend/localsend/pull/669) is set to `1`. This makes GTK properly respect the XDG Decoration protocol.

Additionally, I added [some newer code](https://github.com/flutter/flutter/pull/154522/commits/c99d84f45357986516f3290b50d3af6f0fd08aeb) from the `my_application.cc` Flutter template that makes the title bar icon be correct on KDE Wayland and not a generic Wayland icon.

Another small benefit is that the [Unite shell extension](https://extensions.gnome.org/extension/1287/unite/) for GNOME can merge the regular GTK title bar with the top bar, but not header bars.

| Environment | Before | After |
|-|-|-|
| GNOME | GNOME before | GNOME after |
| KDE |KDE before |  KDE after |
| Sway | Sway before | Sway after |

Note that I also had to upgrade the `path` Dart dependency from `1.9.0` to `1.9.1` to actually build the app but that is not included in this PR.

### DIFF
--- a/app/linux/my_application.cc
+++ b/app/linux/my_application.cc
@@ -23,28 +23,10 @@ static void my_application_activate(GApplication* application) {
   GtkWindow* window =
       GTK_WINDOW(gtk_application_window_new(GTK_APPLICATION(application)));
 
-  // Use a header bar when running in GNOME as this is the common style used
-  // by applications and is the setup most users will be using (e.g. Ubuntu
-  // desktop).
-  // If running on X and not using GNOME then just use a traditional title bar
-  // in case the window manager does more exotic layout, e.g. tiling.
-  // If running on Wayland assume the header bar will work (may need changing
-  // if future cases occur).
-  gboolean use_header_bar = TRUE;
-#ifdef GDK_WINDOWING_X11
-  GdkScreen* screen = gtk_window_get_screen(window);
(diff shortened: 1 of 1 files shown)


landsman:
@nixigaj thank you

landsman:
@Tienisto can you please release only this fix for debian users as a minor release?

faizul726:
> @Tienisto can you please release only this fix for debian users as a minor release?

It should be available on Flathub and other releases now (v1.18.0) @landsman

Issue #2373, opened 2025-03-24
Bug report -- LocalSend Blocks Mac Screen Saver and Sleep Mode for Too Long


Evan-Katz-Benoni-Capital:
### Describe the bug

I have identified, confirmed and reproduced a clear bug in LocalSend.

Initially, it should be mentioned that LocalSend prevents Mac sleep mode when sending a file/folder, which I understand and is totally ok.

However, I have confirmed that LocalSend once again allows Mac sleep mode only when the user clicks on “Done”, on the sending/source Mac, after the file/folder has been sent.  So, if the user is away from his/her Mac, it will not enter screen saver nor sleep mode.  That is, if the user has told his/her Mac to lock the screen after, say, 5 or 10 mintues, LocalSend will prevent sceen lock forever, which is a huge security hole/issue.

Accordingly, the fix for the above bug is for LocalSend to turn off screen lock and sleep prevention (i.e., to once again allow Mac sleep mode) as soon as the file/folder transmission is complete, even before and if the user has not yet clicked on “Done”.  :) 

You can read a write-up of this bug and fix at this link:  https://chatgpt.com/share/67e05011-523c-8007-acb8-e9a5ced9bf32   Go to the very end of the ChatGPT session for the LocalSend bug and write-up.

### To reproduce

1.  Send a file with LocalSend.
2. Do *not* click on "Done".
3. Observe how the sending Mac computer no longer will enter screen saver nor sleep mode.  Huge security hole/issue!
4. You also can see in Activity Monitory how the LocalSend process (in Activity Monitor "Energy" tab) now shows "Yes" regarding "Preventing Sleep".  This is why the sending Mac no longer will enter screen lock nor sleep.
5. In LocalSend, on the sending Mac, now click on "Done".  And you will see in Activity Monitor that LocalSend now says "No" for "Preventing Sleep".
6. So, the fix/solution is to have LocalSend turn "Prevening Sleep" back to "No" immediately after the file is transmitted, without waiting for the user to click "Done".

### Expected behavior

Please see above discussion.  Thank you.

### Screenshots

<img width="763" alt="Image" src="https://github.com/user-attachments/assets/91b6c5da-4dbe-4525-94f8-a6a83f3baaa5" />
<img width="753" alt="Image" src="https://github.com/user-attachments/assets/a043b8ee-ce59-497f-aac0-40ff8625e9f8" />

### Desktop (please complete the following information)

macOS Sequoia, current release v. 15.3.2

[...]

PR #2457, opened 2025-05-13
fix: #2373 disable sleep preventing when files are uploaded/downloaded

ivansapr:
#2373 
Add check if uploading is finished, to disable wake lock.

Straightforward solution with minimum changes.


Tienisto:
Good catch

PR #2493, opened 2025-06-01
fix: delete partial files on transfer cancellation

OldStager01:
# Fix: Delete Partial Files on Transfer Cancellation

## Summary

This pull request introduces logic to automatically clean up partially transferred files when a transfer session is canceled—either by the sender or receiver. This ensures that incomplete or failed files do not persist on the device, improving overall system reliability and user experience.

## Changes Introduced

- **Partial File Cleanup**
  - Added a `_cleanupPartialFiles()` method to delete files with `FileStatus.sending` or `FileStatus.failed`.
  - Included error handling and logging to track cleanup results.

- **Improved Cancellation Handling**
  - Integrated cleanup logic in both sender-initiated and receiver-initiated cancellations.
  - Updated session state to reflect `canceledBySender` status and set the appropriate `endTime`.

- **Accurate File Path Assignment**
  - Updated the file path during transfer initiation using:
    ```dart
    path: saveToGallery ? null : destinationPath
    ```
    - This ensures the actual destination file name is captured (including any auto-renamed variants like `file(1).txt`), allowing accurate reference and cleanup if needed.
    - Note: The path is also updated again upon successful transfer (existing logic).

- **State Management**
  - Ensured the server’s session state remains consistent, especially when file paths or statuses are modified mid-transfer.

## Rationale

- Without cleanup, canceled transfers could leave behind partial or corrupted files, leading to wasted storage and potential user confusion. 
- Additionally, capturing the final resolved file path early allows proper handling of renamed files due to name collisions.

## Testing

- Manually tested various cancel scenarios during file transfer.
- Verified that:
  - Partial files are deleted appropriately.
  - Session metadata and UI reflect correct cancellation state.
  - File paths are correctly resolved and used for cleanup.

---

This update enhances transfer reliability by preventing orphaned files and ensuring consistent state management across cancellation scenarios.

### DIFF
--- a/app/lib/provider/network/server/controller/receive_controller.dart
+++ b/app/lib/provider/network/server/controller/receive_controller.dart
@@ -461,7 +461,7 @@ class ReceiveController {
             files: {
               ...oldState.session!.files,
               fileId: oldState.session!.files[fileId]!.copyWith(
-                path: saveToGallery ? null : destinationPath,
+                path: destinationPath,
               ),
             },
           ),
@@ -798,7 +798,6 @@ void _cancelBySender(ServerUtils server) {
     Routerino.context.popUntil(ReceivePage);
   }
 
-  // Add cleanup for partial files
   // Clean up any partially transferred files
   _cleanupPartialFiles(receiveSession);
 
@@ -814,7 +813,7 @@ void _cancelBySender(ServerUtils server) {
   server.ref.notifier(progressProvider).removeSession(receiveSession.sessionId);
 }
 
-// Add a new method to clean up partial files/// Cleans up partially transferred files for a given receiving session
+// Cleans up partially transferred files for a given receiving session
 void _cleanupPartialFiles(ReceiveSessionState receiveSession) {
   for (final receivingFile in receiveSession.files.values) {
     if (receivingFile.status == FileStatus.sending || receivingFile.status == FileStatus.failed) {
--- a/app/lib/provider/network/server/controller/receive_controller.dart
+++ b/app/lib/provider/network/server/controller/receive_controller.dart
@@ -455,6 +455,19 @@ class ReceiveController {
         createdDirectories: receiveState.createdDirectories,
       );
 
+      server.setState(
+        (oldState) => oldState?.copyWith(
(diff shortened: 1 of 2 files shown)


dhgatjeye:
This needs to be added. That's major problem.

IshanJawale:
This needs to be added since I use Local Send Extensively and face this problem a lot of times...

OldStager01:
I've made changes to the code so that all CI tests are now passing. Let me know if anything else needs to be addressed.

app/lib/provider/network/server/controller/receive_controller.dart
```diff
@@ -455,6 +455,19 @@ class ReceiveController {
         createdDirectories: receiveState.createdDirectories,
       );
 
+      server.setState(
```
Tienisto, line 458: Why did you add this? Even if `saveToGallery` is `true`, the file needs to be saved to a temporary path first.

app/lib/provider/network/server/controller/receive_controller.dart
```diff
@@ -784,12 +798,40 @@ void _cancelBySender(ServerUtils server) {
...
           status: SessionStatus.canceledBySender,
           endTime: DateTime.now().millisecondsSinceEpoch,
         ),
       ));
+
+  // Update UI to reflect cancellation
+  server.ref.notifier(progressProvider).removeSession(receiveSession.sessionId);
+}
+
+// Add a new method to clean up partial files/// Cleans up partially transferred files for a given receiving session
```
Tienisto, line 817: Please remove the first comment part

app/lib/provider/network/server/controller/receive_controller.dart
OldStager01: You're absolutely right! Even for gallery files, the temporary cache path is needed for cleanup during transfer. Setting `path: null` breaks the cleanup logic since it can't find partial files.

Just tested this, partial gallery files weren't getting cleaned up on cancellation. 

Will change to:
```dart
path: destinationPath,
```
Thanks for catching this!

PR #2661, opened 2025-08-19
feat: enable starting text share via command line

guilhermetk:
## Summary

• Add support for starting text share via command line using --text or -t flag
• Enables users to quickly share text content by passing it as a command line argument

## Changes

• Enhanced LoadSelectionFromArgsAction in selected_sending_files_provider.dart to handle --text and -t flags
• Added logic to process text content passed after the text flag and add it as a message for sharing
• Maintains existing file sharing functionality while extending command line capabilities

This feature improves the CLI experience by allowing direct text sharing without needing to launch the full UI first.

### DIFF
--- a/app/lib/provider/selection/selected_sending_files_provider.dart
+++ b/app/lib/provider/selection/selected_sending_files_provider.dart
@@ -289,9 +289,14 @@ class LoadSelectionFromArgsAction extends AsyncReduxActionWithResult<SelectedSen
   Future<(List<CrossFile>, bool)> reduce() async {
     bool filesAdded = false;
     bool nextShare = false;
+    bool nextText = false;
     for (final arg in args) {
       if (arg == '--share') {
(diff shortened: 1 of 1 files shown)


ZihaoFU245:
This would be a great feature as it can work on remote Linux servers. How is the progress now?

guilhermetk:
The functionality you mention is not what this PR proposes. It actually allows the GUI to be opened directly to the text sharing screen with the content passed as the parameter value, you still have to select the destination device. 
There is a cli under development, by the maintainer team, that might allow the behavior you expect.

ShlomoCode:
Can't this be achieved using `open` command and it's equivalents?

guilhermetk:
I did try a bunch of variations: open, send... but couldn't manage to open the GUI with a predefined text to be shared, that's why I created this PR.

PR #2676, opened 2025-09-02
feat(android): quick launch tile

Voltra:
## Introduction

This PR introduces a [quick settings tile](https://developer.android.com/develop/ui/views/quicksettings-tiles) to launch the app from the quick settings panel on Android.

## Further Development

- Temporarily disable the tile while a transfer is in progress
- Noop if the app is already on the foreground


Voltra:
I'd like to `Up` this PR

This is a nice QoL improvement (kinda like Shazam's quick scan settings tile)

If you have any questions, feel free to ask

Tienisto:
Nice!

Voltra:
Is there an ETA for it being published as an update on the stores?

PR #2686, opened 2025-09-04
Set WindowClass

xmha97:


Tienisto:
Can you explain why this change is needed?

PR #2692, opened 2025-09-06
Update tr.json

baltrun:
There was one thing wrong, I changed it.


erentomurcuk:
As written in #2689 this string has been replaced with below

image

Reason being this is written as "Panodan Yapıştır" is because this is a button that enables pasting & sending whatever is available in the clipboard.

FYI

Issue #2359, opened 2025-03-16
Bug report: Client can not download files when using "Share via link" from Android


Tudor3510:
### Describe the bug

Clients can not download the files from the webpage generated via "Share via link" from an Android device.

### To reproduce

1. Go to the webpage generated via "Share via link" from an Android device.
2. Accept the request on the Android device.
3. In browser, click on the file you want to download.
4. Wait indefinitely for file to start downloading.

### Expected behavior

The file should start downloading when you click on it

### Screenshots

![Image](https://github.com/user-attachments/assets/5488339f-4472-4422-99c0-3f1900cc617c)
![Image](https://github.com/user-attachments/assets/82d812fa-1d64-40f5-8670-59a7aa93e25f)

### Smartphone (please complete the following information)

SM-J500F with custom ROM (ViperOS with Android 7.1.2)
Xiaomi Redmi Note 8 Pro (with default ROM, Android 11)

### Additional context

Tried downloading from different clients: Chrome (on Ubuntu and Windows), Firefox (on Ubuntu and Windows), Edge on Windows, Safari on iOS. Every time you click on a file in the web interface, the downloading is not starting.

Zhenchik88:
Same problem here.

Nicollas1705:
I faced the same problem, but could bypass it by sharing (to LocalSend) using another app (like a file manager)

Zhenchik88:
> I faced the same problem, but could bypass it by sharing (to LocalSend) using another app (like a file manager)

Sharing from another app really WORKS!
Thanks, Nicollas!

Oppzippy:
Here are logs if it helps:

```
05-08 22:39:28.461  9645 11154 I flutter : 2025-05-08 18:39:28.461205    [INFO] [Server] Starting server...
05-08 22:39:28.463  9645 11154 I flutter : 2025-05-08 18:39:28.463451    [INFO] [Server] Server started. (Port: 53317, HTTP only)
05-08 22:39:34.236  9645 11154 E flutter : [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: PathNotFoundException: Cannot retrieve length of file, path = 'content://com.android.providers.downloads.documents/document/raw%3A%2Fstorage%2Femulated%2F0%2FDownload%2Ftest.txt' (OS Error: No such file or directory, errno = 2)
05-08 22:39:34.236  9645 11154 E flutter : #0      _File.throwIfError (dart:io/file_impl.dart:675)
[...]

PR #2756, opened 2025-10-12
fix: android shared links not downloading files

ShlomoCode:
Fixes https://github.com/localsend/localsend/issues/2359 (regression introduced in https://github.com/localsend/localsend/pull/2044)
This also saves unnecessary duplication of the file into memory (`tempFile`)


Green392:
It is still not working for me. Maybe it broke again? Please look into this once more.


I just realised something. The app hasn't been updated for a whole year. 
So will it ever get an update anytime soon to include this fix?

Issue #2516, opened 2025-06-17
Feature request


66my:
### Is your feature request related to a problem? Please describe.

I often use Android to transfer pictures and other batch files to PC, so I rely on the "select" operation. The select operation feature of Android album can realize "drag and select". Long press and slide to select all files or pictures from the initial press to the lift. I hope to be compatible with this function. Thank you

### Describe the solution you'd like

The select operation feature of Android album can realize "drag and select". Long press and slide to select all files or pictures from the initial press to the lift. 

### Describe alternatives you've considered

I hope to be compatible with this function. Thank you

PR #2769, opened 2025-10-15
feat: add drag-to-select for media batch selection

ShlomoCode:
Closes https://github.com/localsend/localsend/issues/2516


Tienisto:
The attribute should be already enabled by default as per documentation of the function.

ShlomoCode:
I couldn't drag in v1.17 so I thought it was turned off, it seems it was just added in a later version of `flutter_wechat_assets_picker`, meaning it was enabled thanks to https://github.com/localsend/localsend/commit/430586897075174fe999e7de2528cca39455ed8c which has not yet been released

Tienisto:
that's weird because I can swipe. This feature should be available since 9.5.0 that is used by LocalSend in v1.17.0

ShlomoCode:
> that's weird because I can swipe

Wow, that's weird

Issue #2904, opened 2026-01-09
bug: Receive Text or Paste are incorrectly identified as link when message starts with URI but has more text after


numericOverflow:
### Read [this](https://github.com/localsend/localsend/issues/528) before anything

- [x] I have read it.
- [x] Issues not fitting are closed.

### Make sure about:

- [x] This issue is not duplicated with any other open or closed issues.
- [x] I am using the latest version from the release.

### Describe the bug

Text/Paste incorrectly being identified by receiver as links whenever the text *begins* with a URI/URL

### Expected behavior

Text should *ONLY* be identified as a link if the **entire** thing is a link (not when there's other non-URL text in message.  

All of these text strings should be identified as "text" and not "link" (Currently these FAIL):
- `https://gitlab.com has link as starting text so this will be misidentified as link`
- `https://duckduckgo.com is better than https://google.com - this will be misidentified as one big link`

These work OK:
Seen as link:
- `https://duckduckgo.com`

Seen as text: 
- `whereas https://duckduckgo.com with link not at beginning is correctly seen as text`

When there is extra text beyond just a URL and Localsend incorrectly classifies the entire text as a link which it then lumps together as an unusable link.

## On Sending device:
Go to either of these:
- Send > Text
- Send > Paste

Send either of these text strings:
- `https://gitlab.com has link as starting text so this will be misidentified as link`
- `https://duckduckgo.com is better than https://google.com - this will be misidentified as one big link`

## On Receiving device:
Received messaged are incorrectly marked as a "link".  This shows an "Open" button that will attempt to pass the entire string as a URL to web browser, even though there's more text than just the URL start it matched to make it think the text was a URL.

### Device information

Tested it is bug on following configurations
Android v1.17.0 --SEND-TO--> Windows v1.17.0 
Windows v1.17.0  --SEND-TO--> Windows v1.17.0 

### Screenshots

<img width="556" height="444" alt="Image" src="https://github.com/user-attachments/assets/096a93fe-d075-46ac-bd7c-0ea9e18d26b0" />

### Additional context

I'm guessing there must be some issue with the message parsing on receiver side that finds a URL pattern match at start of message string and doesn't check to see there is more non-url text which should actually prevent classifying the message as a link.

Some white space before after the link should be OK if trimmed, but it is currently far to permissive when matching a URL

PR #2907, opened 2026-01-12
fix(receive-page): correct isLink detection for messages starting with URL (#2904)

zunohoang:
**Problem:**
Messages starting with a valid URL but containing extra text were incorrectly marked as links.
Example: `https://google.com hello world`

**Cause:**
`Uri.tryParse(message)?.isAbsolute` marks any string starting with a URL as absolute, even with extra text → false positives.

**Fix:**

* Added `isBrowserUrlStrict` for strict URL validation.
* Updated `ReceivePageVm` to use it:

```dart
isLink = message != null && isBrowserUrlStrict(message!)
```

**Result:**

* Only fully valid URLs are detected as links.
* Messages with extra text are no longer clickable.

**Related issue:** #2904

### DIFF
--- a/app/lib/pages/receive_page.dart
+++ b/app/lib/pages/receive_page.dart
@@ -18,6 +18,7 @@ import 'package:localsend_app/util/ip_helper.dart';
 import 'package:localsend_app/util/native/platform_check.dart';
 import 'package:localsend_app/util/native/taskbar_helper.dart';
 import 'package:localsend_app/util/ui/snackbar.dart';
+import 'package:localsend_app/util/url_helper.dart';
 import 'package:localsend_app/widget/device_bage.dart';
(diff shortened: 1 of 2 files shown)


zunohoang:
@Tienisto pls check

Tienisto:
You are correct that `Uri.tryParse(message)?.isAbsolute` is implemented poorly.
However, there is not only `http`/`http`. It can be any protocol type (for example `ftp`).
Maybe `^\w+:\/\/` would be a better regex.

zunohoang:
@Tienisto  I updated the logic to strictly validate the entire string as a URL, not just the prefix.
It now supports multiple protocols (http, https, ftp, etc.) while avoiding false positives like
`https://google.com hello`.
Please check the commit.

PR #2917, opened 2026-01-25
i18n: Added Uzbek (uz) localization

discoveruz:
Added Uzbek (uz) localization


itsbilolbek:
It would be great to have Uzbek language support to Localsend. Can you add Uzbek language to Weblate too?

Tienisto:
Already has been added

PR #2948, opened 2026-02-22
fix: receive history missing entries on concurrent receives

wyh2001:
## Summary
Fix a race where receive history could drop entries when multiple receives finish close together.

## Reproduction
1. Send two very small files (e.g. text) to a device at the same time.
2. Both received, but History shows only one.

## Fix
1. Serialize receive history writes.
2. Add regression test for concurrent adds.


wyh2001:
Added a commit to replace the handwritten lock. Note that this pulls in a new package `mutex`.

PR #2987, opened 2026-03-25
fix: resolve variable shadowing for remote session ID in sendProvider (#2899)

johansutrisno:
# Description

This PR addresses the `RhttpStatusCodeException` (400/500) encountered during file uploads when using the v2 protocol.

## Root Cause Analysis

The issue was caused by **variable shadowing** in the `startSession` method within `lib/provider/network/send_provider.dart`. 

When the application received a response from the `prepareUpload` request, it declared a new local variable using the same name as the existing local session ID:
```dart
final sessionId = response.response!.sessionId; // Shadowing the local ID from line 64
```
As a result, subsequent calls to `state.updateSession` used this remote ID as the lookup key for the local session state. Since the session was indexed by the local ID, the lookup failed (returned `null`), and the `remoteSessionId` was never correctly updated in the state. 

During the actual file upload process, the application attempted to retrieve the `remoteSessionId`, found it to be `null`, and consequently omitted the mandatory `sessionId` query parameter. This led to the receiver (server) returning a `400 Bad Request` or `500 Internal Server Error`.

## The Fix

- Renamed the shadowed variable to `remoteSessionId` during the response processing phase.
- Ensured the `updateSession` call correctly uses the local `sessionId` as the key while updating the state with the `remoteSessionId`.

## Comparison

| Before | After |
| :--- | :--- |
| `final sessionId = response.response!.sessionId;` (Shadowed) | `final remoteSessionId = response.response!.sessionId;` |
| `state.updateSession(sessionId: sessionId, ...)` (Using remote ID) | `state.updateSession(sessionId: sessionId, ...)` (Using local ID) |

## Evidence

https://github.com/user-attachments/assets/b85738ae-7edc-4cf9-b03e-de5f36ea5e77

https://github.com/user-attachments/assets/5925a956-e09b-4c44-aa60-9828494c9525

## Testing

- [x] Verified that the `remoteSessionId` is now correctly stored in the state after a successful `prepareUpload`.
- [x] Verified that file uploads now include the `sessionId` query parameter.
- [x] Confirmed that the `400/500` error is no longer present during file transfer.

**Fixes #2899**

PR #3013, opened 2026-04-10
fix: clean up upload isolate task lifecycle

B67687:
## Summary
- clean up parent isolate stream subscriptions after terminal errors
- always remove upload cancel tokens on success, failure, and explicit cancellation
- add focused unit coverage for the stream adapter and upload token lifecycle

## Scope
- Keeps the `remoteSessionId` regression out of this PR because #2987 already covers that bug and now includes regression tests.
- Defers upload concurrency tuning and the broader throughput investigation to a separate follow-up.

## Testing
- Added tests for:
  - normal stream completion
  - terminal error completion
  - duplicate terminal signal safety
  - successful upload token cleanup
  - failed upload token cleanup
  - explicit cancel token cleanup
- Validated locally in `common` from a clean Windows checkout:
  - `dart pub get`
  - `dart analyze`
  - `dart test`
- In `app`, `flutter pub get` failed under Flutter `3.41.6` because of dependency resolution (`mockito 5.5.0` vs `flutter_test`-pinned `matcher` / `test_api`), which appears unrelated to this PR.

### DIFF
--- a/common/lib/src/isolate/child/upload_isolate.dart
+++ b/common/lib/src/isolate/child/upload_isolate.dart
@@ -65,6 +65,40 @@ abstract class UriContentStreamResolver {
 
 UriContentStreamResolver? _uriContentStreamResolver;
 
+@visibleForTesting
+Future<void> executeHttpUploadTask({
+  required int taskId,
+  required Map<int, CustomCancelToken> cancelTokens,
+  required Future<void> Function(CustomCancelToken cancelToken) upload,
+  required void Function(IsolateTaskStreamResult<double>) sendToMain,
+}) async {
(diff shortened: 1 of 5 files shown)


B67687:
This is still ready from my side.

I kept it scoped to the isolate lifecycle leaks and left the sender-side `remoteSessionId` bug out because #2987 already covers that path.

If a maintainer is able to take a look, I would also appreciate CI being approved for the fork PR when convenient.

B67687:
Ping — just a friendly check-in. This is still ready on my end if anyone has time to take a look or approve CI. Happy to address any feedback.

PR #3038, opened 2026-04-24
feat: add x86_64 rpm package to release workflow

### LINKED ISSUE #1453: Feature request: Release on Fedora in dnf

localsend is not available to install through dnf on fedora

please make localsend installable on fedora through dnf

using the flatpak

### Additional context

i saw that there is a todo to make localsend binaries available in more linux distributions, but i haven't seen an issue about fedora specifically. i'm creating this issue so others can give thumbs up reactions and developers can see how much interest there is for releasing binaries on fedora specifically.

### LINKED ISSUE #3089: fr: Request Native RPM Packages

- [ ] I have read it.

### Make sure about:

- [x] This issue is a feature request, not a bug report.

### Reason of the change

I need it Because I can install native localsend on Fedora Linux , Like as DEB versions on debian based distros. 

### Description of the change

Can you release RPM Packages for localsend for us. We really need it Because AppImage and Flatpak packages are really annoying. I think it should not be hard ( There is a lot of app That written on flutter and have native RPM Packages ) . So , If it is possible Please release RPM alongside other packages.

### The alternatives you've considered

_No response_

### Additional context

_No response_

basil:
## Summary

LocalSend ships `.deb` Linux packages with each release but does not currently ship an RPM. This PR adds an x86_64 RPM to the release pipeline so each release uploads a `.rpm` asset alongside the existing Linux packages and fixes the previously broken `test_rpm` manual workflow so it builds successfully on modern Fedora.

#### `.github/workflows/release.yml`

- Adds a new `build_rpm_x86_64` job that runs in a `fedora:43` container and uses `flutter_distributor package --platform linux --targets rpm`.
- Adds `build_rpm_x86_64` to the `needs:` array of the release publishing job.
- Downloads the `rpm-x86-64-result` artifact in the release publishing job and uploads it as a release asset named `LocalSend-${VERSION}-linux-x86-64.rpm` with MIME type `application/x-rpm`, mirroring the existing `deb` wiring.

#### `.github/workflows/test_rpm.yml`

- Bumps the container from `fedora:38` (EOL) to `fedora:43`.
- Adds `needs: build` so the job consumes the shared `version` output.
- Replaces the hard-coded `/opt/hostedtoolcache/flutter/stable-...` path with `$FLUTTER_ROOT`, which `subosito/flutter-action` sets to the correct location inside the container. This removes a brittle path assumption that had broken over time.
- Removes dead debugging steps (`which flutter`, `which dart`, `PATH` echo).
- Adds Rust toolchain setup (to match the `deb` job, since the app now links a Rust HTTP client).
- Drops `sudo` from the `dnf install` line, as Fedora container jobs already run as root on GitHub-hosted runners.

## Why the inline `sed` patches?

`flutter_app_packager` 0.6.5 has two bugs that prevent `rpmbuild` from completing on Fedora 41+. Both are patched in a dedicated step, documented with comments in the workflow. The short version:

1. RPM 4.20 (Fedora 41+) invokes `%install` from a new `%{_builddir}/%{name}-%{version}-build/` subdirectory. The spec template in `make_rpm_config.dart` uses relative paths like `cp -r %{name}/*` that were written against the old layout where `%install`'s CWD was `BUILD/` itself. The patched paths now prefix each source with `../` so they resolve to the real `BUILD/` parent. (Using `%{_builddir}/%{name}/*` doesn't work because `%{_builddir}` itself was redefined in RPM 4.20 to include the `-build` subdir.)
2. `flutter_app_packager` runs `patchelf --set-rpath '$ORIGIN'` on plugin `.so` files but gates it on the rpath containing `/home` (a workaround for flutter/flutter#65400). GitHub-hosted runners check out under `/__w/...`, so the gate never triggers and plugin `.so` files ship with build-tree rpaths. RPM `check-rpaths` then rejects the package with `ERROR 0002: invalid runpath`. The patch makes `patchelf` always run regardless of the original rpath.

After editing the sources, `flutter_distributor` is deactivated and reactivated so its AOT snapshot is rebuilt against the patched dependency.

Fixing these issues upstream is deliberately deferred to keep this PR focused on unblocking the RPM release. It would be desirable for someone who has the time and interest to follow up to fix the upstream packaging logic.

## Testing Done

- [x] Triggered the "Build RPM package" workflow manually (`workflow_dispatch`) and confirmed the artifact was produced and installable on Fedora 43.
- [x] Installed the produced `.rpm` in Fedora 43 and launched LocalSend.
- [x] Ran `rpm -qlp` on the artifact to sanity-check the file layout.

Fixes #1453
Fixes #3089

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -268,6 +268,95 @@ jobs:
           name: deb-arm-64-result
           path: ${{ steps.find_deb.outputs.deb_path }}
 
+  build_rpm_x86_64:
+    needs: build
+    runs-on: ubuntu-24.04
+    container: fedora:44
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Install dependencies
+        run: dnf install -y clang cmake gtk3-devel ninja-build libappindicator-gtk3-devel jq findutils which git patchelf rpm-build
+
+      - uses: subosito/flutter-action@v2
+        with:
+          flutter-version: ${{ env.FLUTTER_VERSION }}
+          channel: "stable"
+
+      - name: Configure safe directory for Git
+        run: git config --global --add safe.directory "$FLUTTER_ROOT"
+
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: ${{ env.RUST_VERSION }}
+
+      - name: Check rust-toolchain.toml
+        working-directory: app
+        run: rustup show
+
+      - name: Enable dart_distributor
+        run: dart pub global activate flutter_distributor
+
+      # flutter_app_packager 0.6.5 has two bugs that prevent the rpm build from
+      # succeeding on modern Fedora:
+      #
+      # 1. rpm 4.20 (Fedora 41+) runs %install from a new
+      #    %{_builddir}/%{name}-%{version}-build/ subdirectory. The generated
+      #    spec uses relative paths like cp -r %{name}/* that were written
+      #    against the old layout where %install's CWD was BUILD/ itself, so the
+      #    cp now looks one level too deep and fails with "No such file or
+      #    directory". The fix is to prefix each copy source with ../ to reach
+      #    the real BUILD/ parent. (The obvious alternative,
+      #    %{_builddir}/%{name}/*, does not work because %{_builddir} itself was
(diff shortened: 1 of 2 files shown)


basil:
@Tienisto Are you interested in reviewing this PR?

Issue #241, opened 2023-03-05
Help please, I can't sync my phone, hacker is using my phone, down below is the error from Google cloud, I can't verify my own device


rocky19888:
For security, you can no longer access your encrypted data on this device. Try again using a device you’ve recently used to sign in to your Google Account.

Visit g.co/OnDeviceEncryption to learn more.

Tienisto:
What is this? A bug report? A feature request?

Without giving exact details (version, operating system, hardware details), nobody can help you.

PR #3098, opened 2026-06-01
feat: hashtag mode scans the full /16 around each local interface (+ 3 fixes for #2956 regressions)

morefreeze:
This PR adds the hashtag wide-subnet scan feature, plus three fixes for regressions introduced by [`90de693a`](https://github.com/localsend/localsend/commit/90de693a) ("feat: integrate rust http client") which currently block all HTTPS-mode flows between any two LocalSend builds since that commit.

> **The three fix commits are independent of the feature.** If preferred, they can be cherry-picked first and the feature reviewed separately afterward — the feature commit doesn't depend on them in code, only in practice.

> **A note on IP addresses below.** All IPs used in examples (`10.23.45.67`, `10.23.89.241`, etc.) are illustrative — they're not real addresses from any test environment.

---

### Problem
`DeviceType` in `core/src/model/discovery.rs` is tagged `#[serde(rename_all = "SCREAMING_SNAKE_CASE")]`, so the rust v2/v3 register client expects `"deviceType":"DESKTOP"`. The actual LocalSend protocol — and every existing receiver including stable 1.7.0 — emits the lowercase form `"desktop"`. Result: `register` always fails with "error decoding response body" against any released receiver. Direct IP, hashtag, and v3 nonce paths are all blocked.

### Fix
Align the enum with `ProtocolTypeV2`'s lowercase rule (same file, two lines above) and update the test fixtures that hard-coded the uppercase form.

| | Before | After |
|---|---|---|
| Enum serde rule | `SCREAMING_SNAKE_CASE` | `lowercase` |
| Decodes `"desktop"` from 1.7.0 receiver | ❌ | ✅ |

---

### Problem
`reqwest::tls::TlsInfo` is silently empty when combined with `use_rustls_tls()` + `danger_accept_invalid_certs(true)`. The internal NoVerifier path doesn't store the peer certificate chain on the connection, so `response.extensions().get::<TlsInfo>()` returns `None` and every HTTPS register fails with "TLS info not found".

A second, related bug: even when the cert *is* captured, reqwest's connection pool reuse and rustls's session-ticket resumption both skip the Certificate message on subsequent requests to the same peer, so the verifier never runs and the task-local stays empty.

### Fix
Replace the broken `TlsInfo` indirection with a real `rustls::client::danger::ServerCertVerifier` (`CapturingVerifier` in `core/src/http/client/mod.rs`) that accepts any cert and writes the leaf DER into a `tokio::task_local`. A new `with_cert_capture(future)` helper scopes the holder for the lifetime of one HTTPS request; `verify_cert_from_res` reads from the task-local instead of the response extensions. Every v2 and v3 request method that calls `verify_cert_from_res` is now wrapped (`register`, `prepare_upload`, `upload`, `nonce`).

To make the verifier run on every request, `rustls::client::Resumption::disabled()` is set on the `ClientConfig` and `pool_max_idle_per_host(0)` is set on the reqwest builder — forces a fresh handshake per request. The handshake cost is acceptable for LocalSend's request rate; the correctness win is required.

---

### Problem
`app/lib/provider/network/send_provider.dart:243-251` shadows the outer local `sessionId` with an inner `final sessionId = response.response!.sessionId`. `state.updateSession(sessionId: sessionId, ...)` then uses the inner *remote* id as the map key — which doesn't exist in the local session map — so `updateSession` is a no-op, `remoteSessionId` is never stored, and every subsequent `/upload` request omits the `sessionId` query param. The receiver's `_uploadHandler` (`app/lib/provider/network/server/controller/receive_controller.dart:466`) returns `400 "Missing parameters"`.

End-user symptom: file send completes the byte stream then fails at the end with `[RhttpStatusCodeException] Status code: 400. URL: https://<peer>:53317/api/localsend/v2/upload`, regardless of file type or size.

### Fix
Rename the inner binding to `remoteSessionId` so the outer local `sessionId` remains the map key and the new server-issued id is actually stored on the local state.

```dart
fileMap = response.response!.files;
final remoteSessionId = response.response!.sessionId;  // was: final sessionId = ...
state = state.updateSession(
  sessionId: sessionId,                                 // outer (local), now unambiguous
  state: (s) => s?.copyWith(
    remoteSessionId: remoteSessionId,
  ),
);
```

---

### Problem
When the receiving device shares your `/16` but sits on a different `/24` — typical for corporate WiFi/Ethernet on a `/21` or `/20` subnet — typing `#241` from `10.23.45.67` couldn't reach `10.23.89.241`, because hashtag mode only probed the local `/24` (`10.23.45.0/24`). On large corporate networks this rendered hashtag mode useless across VLANs.

### Fix
Extend hashtag mode to search outward from the local third octet, prioritising the actual subnet (read from the OS via `if-addrs`) and falling back to the rest of the `/16`.

**Subnet detection.** A new Rust function `list_local_interfaces` (`app/rust/src/api/network.rs`, backed by the `if-addrs` crate which uses `getifaddrs(3)` on Unix/Apple/Android and `GetAdaptersAddresses` on Windows) is exposed through `flutter_rust_bridge` so every non-loopback IPv4 interface — Ethernet, WiFi, VPN — reports its real prefix length. Dart's `NetworkInterface.list()` drops the netmask, and `network_info_plus.getWifiSubmask()` is WiFi-only, so neither covered Ethernet. When the Rust call fails for any reason the provider falls back to `/16`, matching the previous "full third-octet" coverage.

**Two ordered queues per interface.** `_buildWideQueues` (`app/lib/util/two_direction_scanner.dart`) emits one descending and one ascending queue from the local third octet `c`. In-subnet octets are placed first (outward toward the subnet boundary computed by `subnetBoundsForThirdOctet`), then out-of-subnet octets continue outward toward `0` / `255`.

**Worker pool with work-stealing.** `TwoDirectionScanner` drives the scan with 32 workers split proportionally between the two queues (floor 1 per non-empty direction via `allocateWorkers`). Each worker prefers its own queue; once empty it steals from the other direction. The `ScanItem.source` field records which queue an item actually came from so progress is credited to the queue that drained, even when stolen.

**UI.** The dialog renders two `LinearProgressIndicator` rows (`↓` and `↑`) with `done/total` and a throughput-based ETA (`—` until 5 samples), plus a collapsible "Tried IPs (N)" disclosure. The hint row above shows the detected mask and in-subnet range, e.g. `10.23.45.241 /21 (subnet 40-47)`. `setState` is coalesced at ~20 Hz to avoid jank from 32 concurrent workers. After a failed scan the bars disappear but the tried-IPs list persists, sorted ascending by IPv4 octet.

#### Subnet math (`subnetBoundsForThirdOctet`)

| Prefix | c | (min, max) | Notes |
|---|---|---|---|
| `/16` | any | `(0, 255)` | Full third octet free — same scope as before |
| `/20` | 45 | `(32, 47)` | 16-value range |
| `/21` | 45 | `(40, 47)` | 8-value range |
| `/22` | 45 | `(44, 47)` | 4-value range |
| `/23` | 45 | `(44, 45)` | 2-value range |
| `/24` | 45 | `(45, 45)` | Third octet fully fixed |
| `/25-/32` | 45 | `(45, 45)` | Third octet still fixed |

#### Queue ordering (`buildWideQueues`)

| Interface | `exclude` | Down head | Up head | In-subnet first? |
|---|---|---|---|---|
| `10.23.45.67 /21` | `{10.23.45.241}` | `44, 43, 42, 41, 40, 39, …, 0` | `46, 47, 48, …, 255` | ✅ 44→40 (in-subnet) then 39→0 |
| `10.23.45.67 /24` | `{10.23.45.241}` | `44, 43, …, 0` | `46, 47, …, 255` | n/a — entire wide scan is out-of-subnet |
| `10.23.45.67 /16` | `{10.23.45.241}` | `44, 43, …, 0` | `46, 47, …, 255` | n/a — full /16 covered |

#### Worker allocation (`allocateWorkers`)

| total | downTotal | upTotal | (down, up) |
|---|---|---|---|
| 32 | 45 | 210 | `(6, 26)` — proportional split |
| 32 | 254 | 1 | `(31, 1)` — floor enforced |
| 32 | 0 | 100 | `(0, 32)` — empty side gets nothing |
| 32 | 0 | 0 | `(0, 0)` |

#### Manual verification

- [x] HTTP hashtag wide scan from `10.23.45.67` to `10.23.89.241` (different /24 within /16) — receiver found, dialog pops within ~1s.
- [x] HTTPS direct-IP register works twice in a row (regression test for fix 2's connection-pool path).
- [x] HTTPS hashtag wide scan finds receiver.
- [x] File send over HTTPS (was regression-blocked before fix 3) — full file delivered, receiver reports 200.
- [x] `flutter analyze lib test` — clean.
- [x] `flutter test test/unit/util/two_direction_scanner_test.dart` — 25/25 pass.
- [x] `cargo test --lib --features http` (in `core/`) — 14/14 pass.
- [x] `flutter build macos --release` — succeeds.

### Out of scope

- IPv6 (still filtered out by `local_ip_provider.dart` upstream).
- VPN routing-table awareness (e.g. macOS may route `10.x` via a tunnel whose interface mask doesn't reflect the routable range). The current implementation trusts the interface's mask, which is correct for the vast majority of corporate WiFi/Ethernet setups.
- A manual override for the subnet prefix. Could be added as a setting if requested.


morefreeze:
Hi @Tienisto — apologies for the ping. This PR contains three small fixes that block all HTTPS-mode flows (register / prepare-upload / upload) between any two builds since `90de693a`, plus one larger feature on top. The fixes are independent of the feature, so if it's easier they can be cherry-picked and merged first while the feature gets a longer review. Would appreciate your review whenever you have time. Thanks!

Tienisto:
Thanks for the PR. I will check your PR when I need some ideas.
I won't merge this because the PR is simply too big.

PR #3102, opened 2026-06-03
[experimental] QUIC transport for LAN transfers

TheGB0077:
Hey there! Long time, no see 🤪

Just passing by with a very messy PR! 

Overview: QUIC implementation for the app's transport layer. Potentially as a replacement for the standard TCP over http. 
QUIC has more resilient connection persistence, lower latency for streaming to begin between clients and support for multiple independent streams over a single UDP connection.

what I accomplished on this PR: 

Single stream file transfer with full rust ownsership of the file data, including on Android where using jni was needed to interface with the files api directly. Does it work on IOS/MacOS? idk :P 

Let me know if this seems like a good idea.


TheGB0077:
Doing some Android to linux transfers I have seen some speedups, the issue #384 reports some wide range of speeds, but on my intial testing I already got 35Mb/s from Android to linux in the same-ish conditions as described in the issues

TheGB0077:
Parallel transfers so far work great!

Tienisto:
Nice! Very big PR. Since QUIC is only a replacement for TCP/HTTP, Dart should not know the exact details except that it wants to use QUIC, no?

I've seen you implemented QUIC for LocalSend Protocol v3 as well. This is fine but I think we should also implement v2 first so that we can keep things simple.
V3 is actually never documented yet, only implemented in WebRTC and partially in HTTP but never used in Dart.
It uses ephemal and permanent public keys with pairing (both sides have to agree to share their permanent public keys) - so the protocol is a little bit more complicated.
In V2, you only use the public key as "pairing" (it is more "remembering" than "pairing").

I will check this PR in detail but as I said, most of the logic should be hidden in Rust. So you might need an interface that combines HTTP and QUIC. As of now, I made a big mistake that I used Dart for networking (HTTP server is still in Dart) but every network/protocol related code should be in Rust, so that we can ship a CLI later.

TheGB0077:
Hey Tien! I did register some QUIC variant as V3, but this was a bit of a mistake, due to the implementation hooking into the V2 protocol currently used. With that aside, the current implementation already handles older clients through the standard http + tls handshake, but the next step would be to tackle an interface to handle both providers. 

Gladly though, quic is naturally fit to exist segmented off on the Rust side. In fact, perhaps we could push this even further with https://github.com/n0-computer/iroh? Maybe this would be even a bigger redesign, but I believe this could drastically simplify the stack.

TheGB0077:
iroh as far as I understand implements some RFC features of the standard QUIC address discovery and QUIC NAT traversal. So there is some potential there to just replace WebRTC work as well. At least, I think this would be better than having a separate maintance burden over many standards.

TheGB0077:
For more details:

https://github.com/n0-computer/iroh-blobs

https://github.com/n0-computer/noq

MoRanYue:
I hope this can improve slow transport speed sometimes.

TheGB0077:
I'm taking a quick detour on the iroh implementation. I may end up opening a secondary PR for this as well. It seems to work well enough, but there's quite a bit more to tinker with considering it's a fundamentally different protocol using ticket advertising locally or across networks.
https://github.com/TheGB0077/localsend/tree/feat/iroh-quic

TheGB0077:
been working on some other stuff while waiting on a new router that I ordered. Should be returning to this soon.

Pentaphon:
Any speed tests to compare the two?

TheGB0077:
> Any speed tests to compare the two?

I just added some fixes and got around this speed transfering a big file.

image

TheGB0077:
hey @Tienisto I rebased the PR so I believe you can now test this on your end! I was thinking if given the "experimental" status the option to use QUIC should be enabled via configs? what do you think?

Tienisto:
Thanks for the changes, I am trying to publish the next release a soon as possible. For the next release, I have 2 features in mind:
- speed should be no more of an issue (networking is almost fully offloaded to Rust and will be finished in the next days)
- the http server operates in a separate thread so that we can run LocalSend (at least on Android) in the background

Other features (including QUIC) have lower priority for the next release. I assume that when networking and file I/O is fully handled by Rust, then the speed issues should be resolved.

I also read about iroh. It is definitely interesting to explore. I envision LocalSend to use multiple protocols (HTTP, WebRTC) anyways, so iroh would be just another "channel".

Some general questions, what LLM did you use for that PR? I find Fable to produce the most concise code but the limits are harsh. After the next release, I will checkout your PR again. I am not 100% sure about dealing with LLM based PRs. I might only merge small PRs from now on; big PRs (like yours) maybe won't get merged but will be feed to my LLM session just to fit my personal taste of code style. Unfortunately, your rebase is not 100% clean but that's my fault of committing directly to the main branch haha.

TheGB0077:
hey @Tienisto, on the note of the rebase, I just noticed some files were not properly handled by Gitbutler when I started rebasing, although I didn't see it at the time. 

About LLMs, this PR was a mix between Deepseek, GLM and me. I actually handled the Rust code more closely and deferred the provider hooking to the LLMs, such as in the quic_send and quic_server providers / unified transfer dispatcher. On the other hand, there's quite a lot of codegen included too by the rust to flutter bridge utility. 

I also believe this PR could be smaller, but at the same time I avoided pushing for a major refactor, due to the http code being isolated in your personal http crate, which has since also been removed as per the latest commits, so I'd say there are a few elements to reconsider about this PR after the latest refactors.

Lastly, about the quinn vs iroh implementation. Both are fundamentally QUIC under the hood, so for QUIC support, they are mutually exclusive. The key difference is that the fine folks at n0 built a fork of quinn that also handles QUIC NAT traversal, thus allowing communication with their relays, but personally I couldn't fully make it work when I was trying it out, and I wasn't getting the same raw speed as the quinn bare implementation, so I set it aside, but it is still a valid path to explore further.

PR #3141, opened 2026-06-20
docs: added german readme

TFJ5183:
Hi, Ive added a german translation for the readme. Ive additional fixed the links from the different files (The links are sorted like the following: Current language, english and then the rest sorted by they own country name (in there language))


Tienisto:
Thanks for the PR. Unfortunately, I've decided to only include English and Chinese (zh-CN) in this repo to capture most audience. Supporting all languages like in the app (50+ languages) seems overkill for a document targeting programmers.

PR #3145, opened 2026-06-21
docs: add discovery troubleshooting tips

archit-goyal:
## Summary

- add Linux firewall examples for allowing LocalSend on port `53317`
- mention VPN/local network settings as a common discovery blocker
- document manual IP sending as a fallback when automatic discovery fails

## Validation

- `git diff --check`

No app build was run because this is a README-only documentation change.


archit-goyal:
Hi, just a gentle check-in on this small README-only troubleshooting update. I kept the scope limited to discovery-related cases that show up repeatedly in issues: Linux firewall rules, VPN/local network blocking, and manual IP fallback. Happy to trim or reword anything if you would prefer this guidance to stay shorter.

Tienisto:
Thanks for the PR. I made a similar commit on main.

PR #3167, opened 2026-07-05
Fix mutually exclusive Quick Save settings

### LINKED ISSUE #3163: bug: Both "Favorites" and "On" can be simultaneously active

- [x] I have read it.
- [x] Issues not fitting are closed.

### Make sure about:

### Describe the bug

In the *Settings* menu, both *Quick Save* and *Quick save for "Favorites"* can be activated at the same time, which translates to both *Favorites* and *On* being simultaneously enabled in the *Receive* menu.

### Expected behavior

The *Settings* menu presents the options as mutually exclusive, for example by using a 3-option drop-down instead of two toggles, or by greying out the *Quick save for "Favorites"* toggle when *Quick save* is enabled as the former is implied by the latter.

The *Receive* menu cannot show *Favorites* and *On* as simultaneously active.

### Reproduction

1. Open the *Settings* menu.
2. Enable both the *Quick Save* and *Quick save for "Favorites"* toggle switches.
3. Open the *Receive* menu.
4. Both *Favorites* and *On* are shown as active.

### Device information

Device: Apple iPhone 7
System: iOS 15.8.8

### Screenshots

Settings menu
Receive menu

### Additional context

_No response_

WhiteJadeWolf:
Fixes #3163

## Summary

- This PR prevents the invalid state where both **Quick Save** and **Quick Save for "Favorites"** can be enabled simultaneously.
- Since **Quick Save** is a superset of **Quick Save for "Favorites"**, the two settings should be mutually exclusive.
- Enabling **Quick Save** now disables and clears **Quick Save for "Favorites"**, preventing a confusing and inconsistent configuration.

## Changes

- Added support for disabling `_BooleanEntry` controls.
- Disabled the **Quick Save for "Favorites"** toggle whenever **Quick Save** is enabled.
- Cleared the **Quick Save for "Favorites"** setting while **Quick Save** is enabled, preventing the invalid state from being persisted.

## Result

- The two Quick Save options are now mutually exclusive.
- The Receive page now correctly reflects the selected Quick Save mode.

## Screenshots

(both off - No Quick Save)
screenshot_2026-07-06_02-27-19

(only Quick Save for Favorites switched on)
screenshot_2026-07-06_02-27-01

(Quick Save switched on - Quick Save for Favorites option disabled / greyed out)
screenshot_2026-07-06_02-25-12

### DIFF
--- a/app/lib/pages/tabs/settings_tab.dart
+++ b/app/lib/pages/tabs/settings_tab.dart
@@ -165,6 +165,7 @@ class SettingsTab extends StatelessWidget {
                       ),
                       _BooleanEntry(
                         label: t.settingsTab.receive.quickSaveFromFavorites,
+                        enabled: !vm.settings.quickSave,
                         value: vm.settings.quickSaveFromFavorites,
                         onChanged: (b) async {
                           final old = vm.settings.quickSaveFromFavorites;
@@ -636,12 +637,14 @@ class _SettingsEntry extends StatelessWidget {
 class _BooleanEntry extends StatelessWidget {
   final String label;
   final bool value;
+  final bool enabled;
   final ValueChanged<bool> onChanged;
 
   const _BooleanEntry({
     required this.label,
     required this.value,
     required this.onChanged,
+    this.enabled = true,
   });
 
   @override
@@ -663,7 +666,7 @@ class _BooleanEntry extends StatelessWidget {
             child: Center(
               child: Switch(
                 value: value,
-                onChanged: onChanged,
+                onChanged: enabled ? onChanged : null,
                 activeTrackColor: theme.colorScheme.primary,
                 activeThumbColor: theme.colorScheme.onPrimary,
                 inactiveThumbColor: theme.colorScheme.outline,
--- a/app/lib/provider/settings_provider.dart
+++ b/app/lib/provider/settings_provider.dart
@@ -165,8 +165,12 @@ class SettingsService extends PureNotifier<SettingsState> {
 
   Future<void> setQuickSave(bool quickSave) async {
     await _persistence.setQuickSave(quickSave);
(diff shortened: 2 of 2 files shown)


nicolandu:
@WhiteJadeWolf Thank you for your work! However, shouldn't the **Quick save for favorites** toggle be greyed out in the *on* position when **Quick save** is active? I think your proposed implementation makes it seem as though quick saving for favorites is off when quick saving is on.

Edit: turning off the **Quick save** toggle should still indeed turn off both toggles.

WhiteJadeWolf:
Thank you for the feedback. My reasoning for clearing Quick Save for "favorites" was that I saw the issue as not only a Ul problem, but also a state management one. Since both options are stored as independent booleans, I wanted to avoid persisting a state where both are true, as that's the combination that originally led to the bug.

my idea of Quick Save was replacing the more specific Quick Save for "Favorites" mode rather than coexisting with it, so I made the settings mutually exclusive in both the Ul and the underlying state.

That said, I see your point about the UX. Keeping Quick Save for "Favorites" visually enabled (but actually disabled) does better communicate that it's included within Quick Save.

But then again... I think its better to keep the favourite option disabled while the superset "quick save" is switched on... Maybe we can add "(for all senders)" beside it to make the option more clear that its the "umbrella" option rather than making the UI change

nicolandu:
While your suggestion would be clear, and while it would indeed be a 1:1 mapping of the underlying state where at most one boolean is active at a time, I believe that a graphical representation (a greyed-out enabled switch) is clearer than a textual description of the fact that [favorites]&nbsp;&subset;&nbsp;[everyone].

In any case, I digress, and the bottom line is that either option would be better than the status quo. If anyone wishes to comment, I'm all ears!

PR #3178, opened 2026-07-10
feat: custom save locations

### LINKED ISSUE #2839: fr: Custom Save Locations Based on File Type

- [x] I have read it.

### Make sure about:

- [x] This issue is a feature request, not a bug report.

### Reason of the change

Currently LocalSend allows users to choose only one destination directory for all incoming files. This is inconvenient because different file types are usually stored in different locations (e.g., PDFs in Documents, music in Music folder, videos in Movies, etc.). Users must manually move files after receiving them, which breaks workflow and reduces efficiency.

### Description of the change

Add an optional feature that allows users to set custom saving rules based on file types.
Example:

- All .pdf, .docx, .txt → save to Documents
- 
- All .mp3, .wav → save to Music
- 
- All .jpg, .png, .mp4 → save to Photos/Videos
- 
- Allow users to define and edit rules and destination folders (preferably per file type or per category).
This feature can be optional and disabled by default, keeping the current behavior for users who prefer a single folder.

### The alternatives you've considered

- Manually moving files after the transfer (not convenient, especially for frequent use).
 
- Using external automation apps (e.g., file managers or system automation tools), but these are not available or consistent across different platforms (Android, iOS, desktop).

- Creating separate receiving folders per transfer session (still requires manual sorting afterwards).

### Additional context

This feature would be especially useful for users who frequently share media, documents, and miscellaneous files, and for those who use LocalSend on mobile devices.
Optional enhancements in the future might include:

- Rule presets by category (Documents / Media / Music / Others)
- Auto-detect file type categories

Nert53:
Closes #2839 implementing new feature.

> [!IMPORTANT]
> Changes made in `app/lib/provider/network/send_provider.dart` (renaming the `sessionId` variable) are important because without them the send process finishes with error. The name of variable is covered up, which result in returning null value. 

## Summary
This PR adds new feature that allow user to choose multiple save locations that are automatically assing based on file type. I added 4 rules - default, documents, images and videos, music (audio)

## Changes
- Added new section in setting for toggling this feature and choosing locations.
- Added icons to sections in settings.
- Added destionation directory field for each receiving file.
- Added required backend logic for this feature.
- Updated title on progress page to be more descriptive (Files sending -> Files sent, ...).
- Updated some translation for this feature (en, cs, de).

## Showcase
Screenshot-save-01
Screenshot-save-02

## Future ideas
Maybe it would be nice to put together more of the mimeType functions used and divide files into more categories. It would get us more locations for saving and also different icons that represent transfered file.

### DIFF
--- a/app/assets/i18n/cs.json
+++ b/app/assets/i18n/cs.json
@@ -122,6 +122,14 @@
       "saveToGallery": "Uložit média do galerie",
       "saveToHistory": "Uložit do historie"
     },
+    "save": {
+      "title": "Ukládání",
+      "saveLocation": "Rozdílné umístění pro různé typy souborů.",
+      "defaultLocation": "Výchozí",
+      "documentsLocation": "Dokumenty",
+      "mediaLocation": "Obrázky a videa",
+      "musicLocation": "Hudba"
+    },
     "send": {
       "title": "Poslat",
       "shareViaLinkAutoAccept": "Automaticky přijímat požadavky v režimu \"Sdílet přes odkaz\"."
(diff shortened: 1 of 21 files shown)


AmirrezaGhiasvand:
Wow this is a great addition.
Looking forward to seeing it get merged.

PR #3181, opened 2026-07-10
fix: rename sessionId to receiverSessionId

Nert53:
Fixed a critical error that prevented the transfer from completing due to an improperly named variable. The name sessionId was obscured, causing a null to be returned.


Tienisto:
Thanks this change has already been added during my refactoring process

Issue #3177, opened 2026-07-10
bug: incorrect sub-second timestamps of transferred files


pjw91:
### Read [this](https://github.com/localsend/localsend/issues/528) before anything

- [x] I have read it.
- [x] Issues not fitting are closed.

### Make sure about:

- [x] This issue is not duplicated with any other open or closed issues.
- [x] I am using the latest version from the release.

### Describe the bug

# Preparation
I used Syncthing to sync one file to Android, with `lastModified=2026-07-10T00:00:00.123456789`.
# Bug
I send the file back from Android LocalSend(1.17.0) to Archlinux aur/localsend-bin (1.17.0-1), the received file is `lastModified=2026-07-10T00:00:00.000123000`.

The sub-second part is weirdly *shifted* three digits.

### Expected behavior

Ideally, LocalSend should set timestamps with nanoseconds resolution.

However, DateTime in Dart seems to be limited to microsecond precision?

Microsecond precision is OK, but the weirdly *shift* must be eliminated, which makes `sorting by filename` and `sorting by lastModified` give different results.

### Reproduction

1. prepare a file and verify the result
   1. Install [Termux](https://play.google.com/store/apps/details?id=com.termux) and `termux-setup-storage`
   2. `touch aa.mp4 -d2019-09-12T12:02:45.123456789`
   3. `stat aa.mp4|grep Modify` -> 2019-09-12 12:02:45.**123456789** +0800
3. Use LocalSend to send it to PC (in my case, Archlinux)
4. verify the timestamp
   1. `LANG=C stat aa.mp4|grep Modify` -> 2019-09-12 12:02:45.**000123000** +0800
5. I've test several files, all shifted three digits; thus, it does not look like the floating-point number precision issue. 

#### Sender
Samsung S25U; LocalSend 1.17.0
#### Receiver
x86_64, Archlinux; LocalSend 1.17.0 (from aur/localsend-bin 1.17.0-1)

0x391F:
It seems that high precision file timestamp on Android and Linux is statx_timestamp stx_btime (birth time, i.e. create time).

PR #3187, opened 2026-07-15
test: cover sub-second timestamp metadata

Chethan616:
Addresses #3177 with regression coverage for the current file-metadata path.

The original branch conflicted because the old PrepareUploadRequestDto and its Dart test were removed from main. This refresh is rebased onto current main and keeps the coverage in the active Rust implementation.

### What is covered

- RFC3339 fractional seconds preserve all nine digits (123456789) when parsed.
- The v2 prepare-upload payload preserves the exact timestamp string.
- An end-to-end v2 upload applies the exact nanosecond timestamp to the received file.
- This specifically guards against the reported 123456789 -> 000123000 digit shift.

### Validation

- rustfmt --check --edition 2021 on all changed Rust files
- cargo test --features http --lib - 53 passed
- cargo test --features http --test v2_server test_upload_applies_file_timestamps - 1 passed
- git diff --check

### DIFF
--- a/packages/core/src/model/transfer.rs
+++ b/packages/core/src/model/transfer.rs
@@ -172,11 +172,12 @@ mod tests {
 
     #[test]
     fn parses_fractional_seconds() {
-        // The Dart implementation sends `DateTime.toIso8601String()` of a UTC
-        // value, which includes fractional seconds: 1970-01-01T00:00:00.500Z.
+        // Preserve all fractional-second digits from the RFC3339 value. The
+        // reported regression shifted the digits in `123456789` by three
+        // places, producing `000123000` on the receiving device.
         assert_eq!(
-            metadata("1970-01-01T00:00:00.500Z").modified_time(),
-            Some(SystemTime::UNIX_EPOCH + Duration::from_millis(500)),
+            metadata("1970-01-01T00:00:00.123456789Z").modified_time(),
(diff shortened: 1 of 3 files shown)


packages/core/src/model/transfer.rs
```diff
@@ -172,11 +172,12 @@ mod tests {
...
-        // The Dart implementation sends `DateTime.toIso8601String()` of a UTC
-        // value, which includes fractional seconds: 1970-01-01T00:00:00.500Z.
+        // Preserve all fractional-second digits from the RFC3339 value. The
+        // reported regression shifted the digits in `123456789` by three
+        // places, producing `000123000` on the receiving device.
         assert_eq!(
-            metadata("1970-01-01T00:00:00.500Z").modified_time(),
-            Some(SystemTime::UNIX_EPOCH + Duration::from_millis(500)),
+            metadata("1970-01-01T00:00:00.123456789Z").modified_time(),
+            Some(SystemTime::UNIX_EPOCH + Duration::from_nanos(123_456_789)),
```
Tienisto, line 180: This will fail on Windows

Tienisto (commented):
Thanks for the PR. Unfortunately, the issue belongs to the sender side right now. I am currently fixing it. The unit tests do not fully catch the issue and should fail on Windows.

PR #3213, opened 2026-07-27
fix(android): apply lastModified to SAF-saved files so timestamps are preserved

sergioperezcheco:
Files received on Android 8+ through the Storage Access Framework (the common path for user-picked folders and SD cards) were not getting the sender's last-modified timestamp. applyFileTimestamps early-returned whenever FileSaveTarget.path was null, and prepareFileSaveTarget returns a content URI + file descriptor (path null) exactly on that SAF path, so every received file kept the current time instead. This is what #3193 (timestamps not preserved on Android) reports, and it is the Android-side half of #3177.

This adds a setLastModified method-channel handler that updates DocumentsContract.Document.COLUMN_LAST_MODIFIED via ContentResolver.update, and calls it from applyFileTimestamps when the target is a content:// URI. The change only touches the SAF branch — plain-path saves still use File.setLastModified/setLastAccessed unchanged.

I made the SAF update best-effort: ContentResolver.update returns the affected row count, and a few providers either reject writes to COLUMN_LAST_MODIFIED or report zero rows, so the handler returns false in that case and the received file is kept with its default timestamp rather than failing the transfer. On the ExternalStorage provider (the typical destination) the update is honoured.

I could not add a unit test for this because it needs a real ContentResolver and SAF document, which only exist on-device; the existing metadata round-trip tests in prepare_upload_request_dto_test.dart already cover the parsing/serializing side, which is unchanged.

Addresses #3193 and the Android side of #3177.

### DIFF
--- a/app/android/app/src/main/kotlin/org/localsend/localsend_app/MainActivity.kt
+++ b/app/android/app/src/main/kotlin/org/localsend/localsend_app/MainActivity.kt
@@ -64,6 +64,7 @@ class MainActivity : FlutterActivity() {
                 "createFile" -> handleCreateFile(call, result)
 
                 "openFileForWriting" -> handleOpenFileForWriting(call, result)
+                "setLastModified" -> handleSetLastModified(call, result)
 
                 "openContentUri" -> {
                     openUri(context, call.argument<String>("uri")!!)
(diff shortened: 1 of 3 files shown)


sergioperezcheco:
Closing — this has been addressed upstream in commit b13bbeb3 (PR for 'fix: timestamp handling on Android'), which moves timestamp handling to the Rust core layer. The new approach applies timestamps through the file descriptor for both Path and Android SAF targets, covering the exact use case this PR targeted. Thanks!

PR #3236, opened 2026-08-05
i18n: improve Hebrew translations

ShlomoCode:


Tienisto:
Thanks for the PR but to be honest, the time part is a bit too complex to me. What about dropping days altogether (dropping `hoursUnitWithAnd`, `daysUnit` and `days`? I mean `32 hours` sounds not bad. I don't think anyone would transfer that long anyways. I think that code path > 24h will be used very rarely

Tienisto:
I will merge this and simplify. Feel free to make a follow up PR if I you have some nitpick

PR #3245, opened 2026-08-07
Ios26 UI

huajingbin:
Add liquid glass tab


app/lib/util/native/device_info_helper.dart
```diff
@@ -13,6 +13,8 @@ Future<DeviceInfoResult> getDeviceInfo() async {
   final String? deviceModel;
   int? androidSdkInt;
 
+  double IOSVersion = 0;
```
GiorgioMendieta: Why double and not int ?

Issue #3165, opened 2026-07-03
bug: "Send to LocalSend" right-click causes port conflict and opens duplicate window


chenyyoung:
### Read [this](https://github.com/localsend/localsend/issues/528) before anything

- [x] I have read it.
- [x] Issues not fitting are closed.

### Make sure about:

- [x] This issue is not duplicated with any other open or closed issues.
- [x] I am using the latest version from the release.

### Describe the bug

When LocalSend is minimized to the system tray, selecting a file in File Explorer and using the context menu "Send to → LocalSend" triggers a second instance of LocalSend to launch. This second instance fails to bind to the default port (53317) because the tray instance already holds it, causing an "Address already in use" error and opening a duplicate main window.

### Expected behavior

The right-click should either:

Pass the selected file(s) via IPC to the running tray instance for sending, or

Bring the existing tray instance to foreground with the file pre-loaded.

No duplicate window or port conflict should occur.

### Reproduction

Launch LocalSend and minimize it to the system tray.

Right-click any file in File Explorer.

Choose "Send to → LocalSend" from the context menu.

A new LocalSend window appears with a port conflict warning / blank state instead of sending the file to the existing tray instance.

### Device information

Windows10 LTSC 21H2

PR #3246, opened 2026-08-07
fix(app): repair the single-instance "show" handover so Send-To reaches the running instance

antheducation:
### What happens today on `main`

On desktop, `preInit` asks an already-running instance to show itself and take over the arguments, and exits if that succeeds. On `main` that request cannot succeed, so every "Send to → LocalSend" starts a second process, which then fails to bind port 53317 and appears as a duplicate window.

Reproduction, on a Windows build of `main`:

1. Start LocalSend and let it bind 53317.
2. In Explorer, right-click any file → **Send to → LocalSend**.
3. A second `localsend_app.exe` starts and stays alive with its own window, while the first process keeps the port.

Windows Explorer creates that second process by design: the Send-To entry is a plain shortcut to `Platform.resolvedExecutable` with no DDE (`app/lib/util/native/context_menu_helper.dart`), so there is no OS-level single-instance guard. The `show` handover is the only guard there is.

### Two root causes, on the same request

**1. The internal `show` endpoint was addressed on the v1 path.**

`preInit` built the URL with `ApiRoute.show.targetRaw(..., peerProtocolVersion)` (`app/lib/config/init.dart:103` on `main`). `peerProtocolVersion` is the constant `'1.0'` (`packages/localsend_isolates/lib/constants.dart:14`), and `targetRaw` maps `'1.0'` to the v1 path (`packages/localsend_isolates/lib/api_route_builder.dart:39`). The server registers exactly one show route:

```rust
// packages/core/src/http/server/mod.rs:540
(&Method::POST, "/api/localsend/v2/show") => internal::show(req, state).await,
```

There is no v1 route in the Rust server, so the handover got a 404, `statusCode != 200`, and the second instance kept starting. `show` is an internal endpoint used to talk to another copy of this same app on this same machine — it is not a peer's negotiated protocol version and should never have been version-negotiated.

**2. No client certificate was presented.**

TLS is on by default (`isHttps()` returns `true` by default, `app/lib/provider/persistence_provider.dart:494`), and the server makes client certificates mandatory whenever the web pages are not served:

```rust
// packages/core/src/http/server/mod.rs:322
let mandatory_client_auth = app_state.web.is_none() && !app_state.web_upload;
```

```rust
// packages/core/src/http/server/common/client_cert_verifier.rs:46
fn client_auth_mandatory(&self) -> bool { self.mandatory }
```

`preInit` used a bare `HttpClient()` (`app/lib/config/init.dart:96` on `main`). `badCertificateCallback` only overrides validation of the *server's* certificate and cannot satisfy a *client* certificate requirement, so the handshake was refused before the request was sent. Confirmed against the real Rust server:

```
no client cert       -> HttpException: TLSV1_ALERT_CERTIFICATE_REQUIRED (ssl/tls_record.cc:486) error 268436572
client cert, v2 path -> 200, and the server emits RsServerEvent.show_(args: [a.txt])
```

Either defect alone produces the duplicate window and the port collision.

### How it got here

Neither commit is wrong on its own reading, which is probably why this passed review:

- `b43b7950` ("fix: remove rhttp", 2026-07-14) replaced `createRhttpClient(<timeout>, persistenceService.getSecurityContext())` with a bare `HttpClient()`, dropping the client certificate. The Dart server at that point did not request client certificates (`HttpServer.bindSecure` without `requestClientCertificate`) and served both `show.v1` and `show.v2` (`app/lib/provider/network/server/controller/receive_controller.dart:115` and `:119`), so nothing broke.
- `1a6d5eae` ("feat(core): add internal show route", 2026-07-15) moved `show` into the Rust server, which serves v2 only and sets `mandatory_client_auth`. Both latent defects went live in that commit.

The token wiring is fine and untouched: `preInit` and the server read the same persisted value (`settings_provider.dart:48` and `server_provider.dart:174`).

### What this PR does

- Adds `packages/localsend_isolates/lib/util/show_instance.dart` with `notifyRunningInstance(...)`. It addresses `ApiRoute.show.v2` directly and gives the client the device's own certificate via `SecurityContext.useCertificateChainBytes` / `usePrivateKeyBytes` from the persisted `StoredSecurityContext`. It lives in `localsend_isolates` because that package owns `ApiRoute` and `StoredSecurityContext`, and because `app` depends only on `localsend_isolates`.
- `app/lib/config/init.dart` calls that helper and exits when it returns `true` (+10 / −32).
- Connect (100 ms) and response (500 ms) timeouts are unchanged; they are named parameters only so the test can widen them.
- The bare `catch (_)` is replaced by outcome-aware logging: a refusal (`statusCode != 200`) and a failed TLS handshake are logged as warnings, because both mean something answered on the port and then denied the handover — the exact shape of this bug. "Nothing is listening", which is the normal case on a cold start, stays at `fine` and prints nothing at the default log level.

Behavior note for reviewers: `preInit` exits only on HTTP 200. That is the same rule the current `main` uses, but it is stricter than 1.17.0, which exited whenever the rhttp call returned without throwing.

### Tests

`packages/localsend_isolates/test/util/show_instance_test.dart` — three cases: the handover succeeds and forwards the arguments; a rejected token does not exit; nothing listening does not exit. The stub server enforces the real server's contract, so reverting either defect alone fails the suite:

| Reverted | Stub answers | Test |
|---|---|---|
| v2 → v1 path | 404 | fails |
| client certificate removed | 401 | fails |
| both | 401 | fails |
| neither | 200 | passes |

Checks run locally with the pinned Flutter 3.41.9: `dart format --set-exit-if-changed lib test`, `flutter analyze` and `flutter test` all pass in both `app` and `packages/localsend_isolates`.

**One honest caveat.** These tests mint a throwaway identity through `generateSecurityContext()`, so `setUpAll` needs the Rust dylib and the tests self-skip without it. `ci.yml` runs `flutter test` on `ubuntu-latest` and does not run `cargo build`, so **they will skip on CI** — the same way the package's existing `cancel_session_repro_test.dart` already does. They do run locally once `cargo build -p rust_lib_localsend_app` has been done. I did not want to trade that skip for a static certificate and private key checked into the repo. Happy to follow whatever you prefer here.

### Evidence

`main` was built for Windows twice from the same tree — once with `init.dart` at upstream, once with the fix — and the real Explorer Send-To path was exercised against each, via the actual `SendTo\LocalSend.lnk` invoked through `ShellExecute` with the test file as its argument.

| Run | Result |
|---|---|
| unfixed `main` | **2** `localsend_app` processes, **2** visible windows, sustained for 15 s. First PID keeps 53317, second cannot bind |
| with the fix | **1** process, **1** window throughout. The second process exits. 53317 never contested |

The surviving instance came to the foreground on the Send page showing `Files: 1, Size: 27 B`, which is the 27-byte test file — so the arguments really reached the running instance. The fix is not just suppressing the second window.

### Scope

This PR fixes the duplicate-instance behavior as it exists on `main` today. It does not claim to explain the original report: #3165 was filed on 2026-07-03 against release **1.17.0** (tagged 2025-02-19), which predates both commits above and takes a different code path, and I could not reproduce the reported symptom on 1.17.0 on Windows 11. So I have left the issue open rather than auto-closing it — please close it if you consider this the same defect, and I am glad to keep digging on the 1.17.0 path if you would rather treat it separately.

### DIFF
--- a/app/lib/config/init.dart
+++ b/app/lib/config/init.dart
@@ -1,5 +1,4 @@
 import 'dart:async';
-import 'dart:convert';
 import 'dart:io';
 
 import 'package:dart_mappable/dart_mappable.dart';
@@ -39,14 +38,13 @@ import 'package:localsend_app/util/native/tray_helper.dart';
 import 'package:localsend_app/util/notification_strings.dart';
 import 'package:localsend_app/util/ui/dynamic_colors.dart';
 import 'package:localsend_app/util/ui/snackbar.dart';
-import 'package:localsend_isolates/api_route_builder.dart';
-import 'package:localsend_isolates/constants.dart';
 import 'package:localsend_isolates/isolate.dart';
 import 'package:localsend_isolates/model/dto/file_dto.dart';
 import 'package:localsend_isolates/model/dto/multicast_dto.dart';
 import 'package:localsend_isolates/rust/api/logging.dart' as rust_logging;
 import 'package:localsend_isolates/rust/frb_generated.dart';
 import 'package:localsend_isolates/util/logger.dart';
+import 'package:localsend_isolates/util/show_instance.dart';
 import 'package:localsend_isolates/util/transfer_notification.dart';
 import 'package:logging/logging.dart';
 import 'package:refena_flutter/addons.dart';
@@ -93,35 +91,15 @@ Future<RefenaContainer> preInit(List<String> args) async {
     // Check if this app is already open and let it "show up".
     // If this is the case, then exit the current instance.
 
-    final client = HttpClient();
-    client.connectionTimeout = const Duration(milliseconds: 100);
-    client.badCertificateCallback = (cert, host, port) => true;
-
-    try {
-      final uri =
-          Uri.parse(
-            ApiRoute.show.targetRaw(
-              '127.0.0.1',
-              persistenceService.getPort(),
-              persistenceService.isHttps(),
-              peerProtocolVersion,
-            ),
-          ).replace(
-            queryParameters: {
-              'token': persistenceService.getShowToken(),
-            },
-          );
-
-      final request = await client.postUrl(uri);
-      request.headers.contentType = ContentType.json;
-      request.write(jsonEncode({'args': args}));
-      final response = await request.close().timeout(const Duration(milliseconds: 500));
-      if (response.statusCode == 200) {
-        exit(0); // Another instance does exist
-      }
-    } catch (_) {
-    } finally {
-      client.close(force: true);
+    final handedOver = await notifyRunningInstance(
+      securityContext: persistenceService.getSecurityContext(),
+      port: persistenceService.getPort(),
+      https: persistenceService.isHttps(),
+      showToken: persistenceService.getShowToken(),
+      args: args,
+    );
+    if (handedOver) {
+      exit(0); // Another instance does exist
     }
 
     // initialize tray AFTER i18n has been initialized
--- a/packages/localsend_isolates/lib/util/show_instance.dart
+++ b/packages/localsend_isolates/lib/util/show_instance.dart
@@ -1,0 +1,90 @@
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:localsend_isolates/api_route_builder.dart';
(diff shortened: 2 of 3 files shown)


antheducation:
Thank you for the review and merge! Happy to follow up if anything shakes out in testing.

PR #3249, opened 2026-08-07
test: release exe

maceip:
Add GitHub Actions workflow to build and release Windows .exe installer.  This runs during CI and uploads the artifact so that users can download a standalone installer without building locally.


Tienisto:
Just look at [release.yml](https://github.com/localsend/localsend/blob/main/.github/workflows/release.yml)

PR #3250, opened 2026-08-07
android: companion-device bluetooth association

maceip:
Use CompanionDeviceManager to observe associated Bluetooth devices and wake the app when a paired device is nearby.  Adds CompanionBridge, LocalSendCompanionService, companion permissions, and a settings UI entry for pairing.\n\nThis eliminates the need for constant BLE scanning — the OS handles proximity detection and starts the app only when a trusted device is present.


Tienisto:
AI slop

PR #3251, opened 2026-08-07
feat: background service boot receiver and keepalive

maceip:
Add BootStartReceiver to restart the app after a reboot or update, ensuring the background service stays active.  Add WindowsSingleInstanceLock to prevent duplicate instances on Windows.  Add foreground service keepalive methods so the app can idle without a persistent notification while still being available to receive files.

Includes:
- fix: init rustlib before persistenceservice and align jvm targets (move RustLib.init() before PersistenceService.initialize(), force all Android subprojects to compile Java/Kotlin to JVM 17)
- feat: replace foreground notification with workmanager + websocket keepalive (replace flutter_foreground_task with WorkManager periodic worker, add WebSocket keepalive provider)
- feat: background service boot receiver and keepalive (BootStartReceiver, WindowsSingleInstanceLock, foreground service keepalive methods)


Tienisto:
Bro, what AI model do you use? The PR contains so much nonsense, I don't think even Claude / ChatGPT would generate this.

PR #3262, opened 2026-08-08
fix: keep accepting connections after a failed accept

ethicnology:
`start_server_with_listener` propagates the error of `incoming.accept()` with
`?`. That completes the branch of the `tokio::select!` in `start_with_port`,
which cancels the whole server — both listeners, IPv4 and IPv6.

But accepting fails for reasons that say nothing about the listener: a peer that
goes away before the handshake completes, or a process that momentarily runs out
of file descriptors. There is no connection limit either, so the descriptor
budget is reachable.

The device then stops receiving **for good**, with nothing telling the
application, until the app is restarted. Reproduced with a lowered descriptor
limit:

```
before saturation : Ok("HTTP/1.1 200 OK")
descriptors exhausted after 245 opens
INFO localsend::http::server: Server stopped on: 0.0.0.0:41777
after releasing them : Err(ConnectionRefused)
```

## What this changes

The loop logs and carries on. Failures that concern only the connection being
accepted are retried at once; the others back off from 50 ms up to 1 s, because
descriptor exhaustion would otherwise spin — the pending connection stays in the
backlog and fails again immediately.

## What this deliberately does not change

No connection limit is introduced. That is a separate discussion.

`libc` is added to `[dev-dependencies]` so the test can lower its own descriptor
limit. It is already in `Cargo.lock` as a transitive dependency, so it costs no
extra compilation, and dev-dependencies do not reach consumers.

*Written with AI assistance

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2202,6 +2202,7 @@ dependencies = [
  "hyper",
  "hyper-util",
  "if-addrs",
+ "libc",
  "lru",
  "pem 4.0.0",
  "percent-encoding",
--- a/packages/core/Cargo.toml
+++ b/packages/core/Cargo.toml
@@ -41,6 +41,12 @@ uuid = { version = "1.24.0", features = ["serde", "v4"] }
 webrtc = { version = "0.14.0", optional = true }
 x509-parser = { version = "0.18.1", features = ["verify"], optional = true }
 
+[dev-dependencies]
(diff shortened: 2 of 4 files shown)


Tienisto:
The comments are extremly verbose (some of them should be included in the PR description, not in the actual code).
But I can fix it later

PR #3264, opened 2026-08-08
fix: reject malformed tokens and oversized SDP from peers

ethicnology:
Two peer-supplied values are read straight off the WebRTC data channel, before
anything about the peer has been verified.

**`decode_sdp` decompresses without a bound.** 678 KiB of zlib expand to
512 MiB — a ratio of 773 to 1 — and the resident set follows.

**`verify_token_with_result` indexes `parts[0..5]`** on a `Vec` collected from
`split('.')`, which panics for any token with fewer than five segments:

```
panicked at packages/core/src/crypto/token.rs:187:87:
range end index 5 out of range for slice of length 1
```

`verify_token_timestamp` also subtracts a peer-chosen timestamp without
`checked_sub`.

## Scope

`webRTCEnabled` is `false` in the app, so this path does not run today. This is
hardening before it is switched on, not an urgent fix — and the panic aborts the
task, not the process, since no profile sets `panic = "abort"`.

## What this changes

The SDP is capped at 1 MiB once decompressed, far above any real session
description. The token segments are taken one by one instead of slicing a
collected `Vec`. The timestamp uses `checked_sub`.

AI assisted

### DIFF
--- a/packages/core/src/webrtc/webrtc.rs
+++ b/packages/core/src/webrtc/webrtc.rs
@@ -1259,14 +1259,34 @@ fn encode_sdp(s: &str) -> String {
     base64::encode(&compressed)
 }
 
+/// The largest SDP that is accepted from a peer, once decompressed.
+///
+/// A session description, ICE candidates included, is a few kilobytes; this
+/// leaves room for an unusually large one while keeping the decompressed size
+/// bounded.
+const MAX_SDP_SIZE: usize = 1024 * 1024;
+
 fn decode_sdp(s: &str) -> Result<String> {
     let decoded_data =
         base64::decode(s).map_err(|e| anyhow::anyhow!("Base64 decode of SDP failed: {e}"))?;
-    let mut d = ZlibDecoder::new(&*decoded_data);
-    let mut result = String::new();
-    d.read_to_string(&mut result)
+
+    // Bounded because the peer chose what to compress: zlib expands by up to
+    // ~1000x, so a payload of a few hundred kilobytes would otherwise be
+    // enough to exhaust this device's memory. The SDP is decoded before
+    // anything about the peer has been verified.
+    let mut decompressed = Vec::new();
+    ZlibDecoder::new(&*decoded_data)
+        .take(MAX_SDP_SIZE as u64 + 1)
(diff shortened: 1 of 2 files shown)


packages/core/src/webrtc/webrtc.rs
```diff
@@ -1373,6 +1393,46 @@ fn to_receive_stream(
...
+    /// An SDP a peer sends is zlib-compressed, and it is decoded before
+    /// anything about that peer has been verified. A payload that expands far
+    /// beyond any real SDP must be refused rather than allocated.
+    #[cfg(target_os = "linux")]
+    #[test]
+    fn test_sdp_decompression_bomb_is_rejected() {
+        // 512 MiB of zeros, which zlib squeezes into a few hundred kilobytes.
+        let bomb = encode_sdp(&"\0".repeat(512 * 1024 * 1024));
+        println!(
+            "charge compressee : {} Kio pour 512 Mio decompresses",
```
Tienisto: This is not English

packages/core/src/webrtc/webrtc.rs
```diff
@@ -1373,6 +1393,46 @@ fn to_receive_stream(
...
+        let pages: u64 = statm.split_whitespace().nth(1).unwrap().parse().unwrap();
+        pages * 4096
+    }
+
+    /// An SDP a peer sends is zlib-compressed, and it is decoded before
+    /// anything about that peer has been verified. A payload that expands far
+    /// beyond any real SDP must be refused rather than allocated.
+    #[cfg(target_os = "linux")]
+    #[test]
+    fn test_sdp_decompression_bomb_is_rejected() {
```
Tienisto: Tracking RAM usage via linux pages is quite hacky. Just leave it out instead of polluting the action runner with 512 MB memory

Tienisto (changes requested):

packages/core/src/webrtc/webrtc.rs
ethicnology: resolved

packages/core/src/webrtc/webrtc.rs
ethicnology: fixed

PR #3276, opened 2026-08-11
feat(cli): replace `--file` with universal `send` command

otanim:
## Summary

- replace `-f` / `--file` with `send [--to TARGET] PATH...`
- accept one or more files, directories, or a mixture in the same transfer
- recursively expand directories while preserving selected root names and relative paths with `/` separators
- select the destination interactively when `--to` is omitted
- optionally resolve an exact, unique device alias with `--to "Device Name"`
- optionally probe an IP address directly over HTTPS on LocalSend port 53317 with `--to IP`
- avoid raw-terminal input and return meaningful failure exit codes in non-interactive mode
- document the complete CLI workflow in the root README and CLI changelog

Empty directories are ignored because protocol transfers contain file entries rather than directory entries. Nested symlinks are not followed.

## Breaking change

The `-f` / `--file` arguments introduced in CLI v1.18.0 are discontinued and replaced by the `send` subcommand.

Before:

```shell
localsend-cli --file report.pdf --file photo.jpg
```

After, with interactive destination selection:

```shell
localsend-cli send report.pdf photo.jpg ./project-backup
```

After, with non-interactive destination selection:

```shell
localsend-cli send --to "Cute Tomato" report.pdf ./project-backup
localsend-cli send --to 192.168.27.26 report.pdf ./project-backup
```

Running `localsend-cli` without a subcommand keeps the existing interactive and receiving mode.

## Automated validation

- `cargo test -p localsend-cli --locked --quiet` (21 passed)
- targeted Rust formatting and diff checks
- targeted CLI Clippy with warnings denied
- `cargo build -p localsend-cli --locked --quiet`
- verified generated top-level and `send` help
- verified `send` rejects an empty path list
- verified a missing `--to` alias exits with status 1 and a clear error

## Live Windows 11 interoperability

- the patched CLI preserved standalone-file behavior and reconstructed a nested folder correctly
- the final `send PATH...` interface sent a mixed selection containing one standalone file and one nested directory as a three-file transfer
- `send --to "Cute Tomato" PATH...` resolved the exact alias and completed after manual receiver acceptance
- `send --to 192.168.27.26 PATH...` directly probed the receiver and completed automatically
- both destination-specific files were confirmed on the Windows receiver
- both `--to` tests ran without the interactive device picker

PR #3277, opened 2026-08-11
feat(cli): add non-interactive target selection

otanim:
Depends on #3276. This is a stacked draft; until #3276 merges, GitHub also shows its base commit and diff. After #3276 merges, this PR will contain only the dedicated target-selection commit.

## Summary

- add optional `send --to TARGET PATH...` destination selection
- resolve an exact, unique device alias from discovery
- probe an IP address directly over HTTPS on LocalSend port 53317
- retain the interactive device picker when `--to` is omitted
- avoid raw-terminal input in non-interactive mode
- return a nonzero exit status when discovery or transfer fails
- document alias and IP usage in the root README and CLI changelog

## Validation

- `cargo test -p localsend-cli --locked --quiet` (21 passed)
- targeted Clippy with warnings denied
- `cargo build -p localsend-cli --locked --quiet`
- live Windows 11 transfer using `--to "Cute Tomato"` (received)
- live Windows 11 transfer using `--to 192.168.27.26` (received)
- missing alias exits with status 1 and a clear error
- both successful tests ran without the interactive device picker

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2256,6 +2256,7 @@ dependencies = [
  "toml",
  "unicode-width",
  "uuid",
+ "walkdir",
 ]
 
 [[package]]
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ To compile LocalSend from the source code, follow these steps:
 
 ## Command Line Interface
 
-The LocalSend CLI is an interactive terminal client built on LocalSend Protocol v2.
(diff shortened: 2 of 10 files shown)


otanim:
Superseded by #3276. The target-selection change has been folded into that PR so the complete `send [--to TARGET] PATH...` replacement remains one cohesive PR and one commit.

PR #3279, opened 2026-08-11
fix(i18n): improve Urdu translations for clarity and consistency

farhaanaliii:
- fix the translation of `on`, it was translated as `پر` which means "on top of"
- make the `copy` translation consistent with other action buttons by changing `کاپی` to `کاپی کریں`
- replace the literal translation of `item` with a more natural word of urdu


Tienisto:
Thanks

Issue #2725, opened 2025-09-26
Surprising behaviour with VPN/tailscale


asdofindia:
I was trying to send a file from linux to android today.

Localsend on linux detected the android device. But when I chose a file and clicked on the device, it failed to show any "accept this file?" message on the android device. 

The problem was that I was connected to tailscale on linux.

The source of the surprise: The android device was being detected — but file was not getting sent. 

Possible solutions:
- This could be a bug which can be fixed thereby allowing to send files even when the sender is connected to VPN.
- It might be unnecessary to support this configuration, in which case, we can add to Troubleshooting section in README

Do let me know if there's any way I can help

theamazing0:
Also experienced this issue, a warning/info message somewhere would be helpful

Issue #2266, opened 2025-02-06
Unable to connect via Tailscale


KlfJoat:
### Describe the bug

I cannot get devices to connect in LocalSend via their Tailscale IPv4 addresses. #1123 has user-submitted anecdotes that LocalSend works when entering the Tailscale IP addresses of other device either as a Favorite or for a targeted send. But that does not work for me.

A different sharing app (KDE Connect) that works much like LocalSend works just fine with Tailscale IPs. Like LocalSend, by default it finds devices with the complementary app via multicast. Also like LocalSend, it allows for specifying devices by IP. But unlike LocalSend, it connects when a Tailscale IP address is entered. So I can guarantee that traffic passes properly between the devices via Tailscale IPs.

I have tried this between 3 devices. My phone as the primary, because it's easy to take off of local Wi-Fi, and my tablet and desktop connected to my home network. I have removed them as Favorites from each other.  I did not have a problem connecting between my tablet and desktop (though there was some weirdness, mentioned in Context, below).

I also tried 2 versions of LocalSend on my Android devices. I started with the F-Droid version, then switched to the Play Store version in case there was a problem in how F-Droid compiled the app.

### To reproduce

1. Disconnect phone from Wi-Fi.
2. Start Tailscale on phone, tablet, and desktop.
3. Install & open LocalSend on all 3 devices.
4. Click on the 'i' icon on a receiving device.
5. On phone (sending device), tap on Send tab.
6. On phone, tap on Text button, enter 'blah' as placeholder, tap Confirm.
7.  On phone, tap on Manual Sending (arrow & target icon).
8. On phone, tap IP Address.
9. Enter the Tailscale IPv4 address of an open LocalSend app.
10. Tap Confirm.
11. Receive message "Error" within 5 seconds.
12. Tap the 'i' icon. Receive message "[RhttpTimeoutException] Request timed out. URL: https://100.1.2.3:53317/api/localsend/v1/info"
13. Go to KDE Connect and make the same connection between the same devices, sending the same 'blah' placeholder text.
14. Repeat steps 6 to 12 to verify that there wasn't some transient network issue that KDE Connect took advantage of or magically resolved.

### Expected behavior

Two LocalSend instances to 'see' each other after I enter the IP address of the other.

### Desktop (please complete the following information)

- OS: Ubuntu 24.10 (Oracular)
- Version: 1.16.1 installed from snap

### Smartphone (please complete the following information)

Phone 
- Device: Android - Google Pixel 7 Pro
[...]

Issue #1123, opened 2024-02-05
Feature request Support Tailscale detection


jdysya:
### Is your feature request related to a problem? Please describe.

As we know, the localsend work in the Same LAN， but [Tailscale](https://github.com/tailscale/tailscale) makes creating software-defined networks easy: securely connecting users, services, and devices. 
Tested by me，it worked when I specify ip address though the app(localsend) can not detect the connection.
So, could we support the  detection ?

### Describe the solution you'd like

I just learned about this tool. I'm not familiar with it.

wkk-dev:
me too

ericxob77:
This would be a great enhancement!  With Tailscale installed localsend could then easily send files to any device located anywhere in the world!

ericxob77:
Please describe in more detail how you got localsend to work through a tailscale connection... I can't figure a way to get it to work.  Thanks!

jdysya:
> Please describe in more detail how you got localsend to work through a tailscale connection... I can't figure a way to get it to work.  Thanks!

You need specify the ip address(in the tailscale) because the app can not detect the connection.
![Screenshot_2024-03-11-12-49-50-263_org.localsend.localsend_app.jpg](https://github.com/localsend/localsend/assets/73117213/21b49b3f-a8ac-406e-9b1f-0149d75fdb3a)

ericxob77:
Aha!  Clicking the arrow pointing to the two concentric circles opens that dialog where one can enter a custom IP address.  Yes, more integration with Tailscale would be cool.

chikei:
the technical difficulties for this feature are 1. tailscale does not support multicast/broadcast yet so there is no efficient way to detect other localsend peer. 2. tailscale assigns ip from 100.0.0.0/8 space so it is not practical (or it is ?) to test all possible ip.

I think currently enter tailscale peer address in localsend's favorites is a good workaround though.

Hk-Gosuto:
I have configured a smaller subnet through access control, which operates normally under localsend.
```json
{
	"nodeAttrs": [
		{
			"target": [
				"autogroup:admin",
			],
			"ipPool": [
				"100.64.200.0/24",
			],
		},
	]
}
```

EDIflyer:
Ooh interesting - agree it's great that the manual IP entry works - although auto-detection would be great, what I've done as a workaround is added those manual IPs for my devices to my favourites list within LocalSend :)

Gdetrane:
[...]

PR #3283, opened 2026-08-11
feat: discover devices over a Tailscale tailnet

hayaksi1:
Devices on the same Tailscale tailnet cannot find each other (#1123). A tailnet carries no multicast,
so announcements are never seen, and its `100.64.0.0/10` range is far past what the `/24` subnet scan
can walk. Every device is directly dialable, yet each stays invisible — the workaround in #1123 is
typing the `100.x` address by hand and saving it as a favourite.

This adds a discovery stage that asks the one process which already knows the peer list — the local
Tailscale daemon — over its LocalAPI, and probes the online peers with the same register request
discovery already sends to a favourite.

|                                       | before                | after                                                                 |
|---------------------------------------|-----------------------|-----------------------------------------------------------------------|
| desktop scan, tailnet peers elsewhere | finds nothing         | probes every online peer the daemon reports (7 of 18 on the test tailnet); the one running LocalSend appears and receives files |
| the phone's list                      | desktop never appears | desktop appears once it has probed or sent to the phone               |
| discovery on a LAN                    | works                 | unchanged — same stages, same timeouts, same escalation               |

## What this changes

- New `packages/core/src/tailscale/` behind a `tailscale` Cargo feature (pulled in by `discovery`).
  It reads the peer list from the local daemon: a unix socket on Linux and on macOS installs running
  `tailscaled`, a named pipe on Windows, the token-guarded loopback port of the sandboxed macOS
  Tailscale builds. Only each peer's address, online flag and display name are used — no key, no node
  identity, and nothing talks to Tailscale's servers. Headscale is the same daemon and works as is.
- Online peers are probed in a stage running beside the announcement. A probe is a register request,
  so a desktop that enumerates the tailnet also becomes visible to the phones — Android and iOS run
  Tailscale as a VPN service with no API for other apps, so a tailnet of only mobile devices still
  cannot discover itself. That limit is the mobile app's, not one this stage can lift.
- LocalSend's own macOS app is sandboxed, which hid the daemon from it entirely: `HOME` names the
  app's container, and the group container holding the loopback credentials cannot be listed. They
  are now read at the two container paths Tailscale ships under, beneath a home taken from the
  password database, and `Runner/*.entitlements` gain a read-only exception for exactly those two
  directories — a temporary-exception entitlement, which App Store review will want justified.
- An inbound registration is the only thing a phone ever gets over a tailnet, so it is never
  discarded: a device sending from a tailnet address is registered on receive, persisted (upsert by
  fingerprint, capped at 16) and re-probed on the first scan after a restart.
- A tailnet address gets a 3 s probe timeout and one retry a second later, where a LAN address keeps
  its 500 ms — first contact has to wake the VPN and punch through NAT or fall back to a relay, which
  on a mobile network regularly outlives a single attempt.
- Two startup-window fixes, not tailnet-scoped because the window exists on any network: a
  registration arriving while discovery is still starting is held (cap 64) and replayed rather than
  dropped, and a scan racing that startup waits briefly instead of being skipped. On a LAN the next
  announcement papered over both; a tailnet has no next announcement.
- Settings → Network (advanced) → "Discover Tailscale devices", on by default. Without Tailscale
  installed the stage costs one failed connect to the local daemon.

## What this deliberately does not change

How devices are found on a LAN. Tailnet confirmations do not count towards the staged-scan
escalation — a peer answering through a tunnel says nothing about whether the local network carries
multicast, and counting it would suppress the subnet scan on exactly the multicast-less networks it
exists for. One knowing ambiguity: `100.64.0.0/10` is shared CGNAT space, so a LAN peer behind such
an ISP classifies as tailnet and is probed with the patient timeout — slower with that peer is the
harmless direction. Only `en.json` is touched; the other locales are Weblate's.

### Tests

Tested on one tailnet between a Linux desktop and an Android phone in both directions, and from a
sandboxed, signed macOS build to that phone, where the desktop probe reached it at its tailnet
address and the transfer completed. `packages/core` gains unit tests for the address classifier,
tailnet channel collection, the probe retry, and the macOS container lookup, plus two integration
tests in `tests/discovery.rs` proving a tailnet-sourced registration does not suppress the
subnet-scan escalation, with a LAN-sourced control. Dart gains a mirror of the classifier test, so
the two classifiers cannot drift apart silently, and an isolate test for the scan/startup race
written against the old code first, where it fails. Not proven: the Windows named-pipe transport is
compile-checked but has not spoken to a live daemon, and phone-to-macOS discovery after a phone
restart has not been observed.

### Scope

Discovery only — transfers over a tailnet already work once devices know each other, which is #1123's
manual workaround. NetBird and ZeroTier run different daemons and are not covered. #2266 and #2725
are adjacent but separate defects and are not claimed here.

### DIFF
--- a/app/assets/i18n/en.json
+++ b/app/assets/i18n/en.json
@@ -147,6 +147,8 @@
       "generateRandomAlias": "Generate random alias",
       "portWarning": "You might not be detected by other devices because you are using a custom port. (default: {defaultPort})",
       "encryption": "Encryption",
+      "tailnetDiscovery": "Discover Tailscale devices",
+      "@tailnetDiscovery": "Setting to also find devices that are reachable over a Tailscale tailnet, but not on the same local network. Tailscale is a VPN product; keep the product name untranslated.",
       "multicastGroup": "Multicast address",
       "multicastGroupWarning": "You might not be detected by other devices because you are using a custom multicast address. (default: {defaultMulticast})"
     },
--- a/app/lib/config/init.dart
+++ b/app/lib/config/init.dart
@@ -167,6 +167,7 @@ Future<RefenaContainer> preInit(List<String> args) async {
             protocol: settings.https ? ProtocolType.https : ProtocolType.http,
             multicastGroup: settings.multicastGroup,
             discoveryTimeout: settings.discoveryTimeout,
+            tailnet: settings.tailnetDiscovery,
             serverRunning: true,
             download: false,
           ),
--- a/app/lib/gen/strings.g.dart
+++ b/app/lib/gen/strings.g.dart
@@ -4,7 +4,7 @@
 /// To regenerate, run: `dart run slang`
 ///
 /// Locales: 58
-/// Strings: 21022 (362 per locale)
+/// Strings: 21023 (362 per locale)
 
 // coverage:ignore-file
 // ignore_for_file: type=lint, unused_import
--- a/app/lib/gen/strings_en.g.dart
+++ b/app/lib/gen/strings_en.g.dart
@@ -1205,6 +1205,11 @@ class Translations$settingsTab$network$en {
   /// en: 'Encryption'
   String get encryption => 'Encryption';
 
+  /// Setting to also find devices that are reachable over a Tailscale tailnet, but not on the same local network. Tailscale is a VPN product; keep the product name untranslated.
+  ///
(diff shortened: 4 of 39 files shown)


Tienisto:
The PR is quite big. What AI model did you use?

hayaksi1:
Fable + GPT-5.6 Sol(as a sidecar)

PR #3300, opened 2026-08-13
fix: avoid response TLS info dependency

### LINKED ISSUE #3299: bug: macOS sends fail with TLS info not found when a packet tunnel is active
### Describe the bug

On macOS, sending an encrypted file to another LocalSend device fails immediately after the receiver accepts when a system packet-tunnel application is active. The reproducible application in this report is Shadowrocket.

The sender displays:

```text
RsHttpClientError.other(field0: TLS info not found)
```

Receiving files on the same Mac continues to work.

### Controlled reproduction matrix

Both LocalSend devices had encryption enabled and LocalSend was restarted after changing the setting.

| Sender | Receiver | Shadowrocket state | Result |
| --- | --- | --- | --- |
| macOS | iPhone | off on both | success |
| macOS | iPhone | on only on macOS | `TLS info not found` after acceptance |
| macOS | iPhone | on only on iPhone | success |
| iPhone | macOS | on on iPhone | success |

Disabling encryption on both devices changes the symptom to a completed transfer followed by HTTP 500, but does not address the encrypted-transfer failure.

### Steps to reproduce

1. Enable encryption in LocalSend on macOS and iOS.
2. Start Shadowrocket on macOS in its packet-tunnel/enhanced mode.
3. Keep both devices on the same Wi-Fi network.
4. Send a file from the Mac to the iPhone.
5. Accept it on the iPhone.

### Expected behavior

The encrypted transfer succeeds while the macOS packet tunnel remains active. Users should not have to disable their normal internet tunnel to transfer a LAN file.

### Actual behavior

The Mac reports `RsHttpClientError.other(field0: TLS info not found)` and the encrypted transfer fails.

### Environment

- LocalSend macOS: 1.18.0 (build 60)
- Receiver: iPhone
- macOS network extension: Shadowrocket packet tunnel / enhanced mode
- LocalSend protocol: HTTPS, v2
The macOS routing table kept the LAN route on Wi-Fi while the default route used a `utun` packet-tunnel interface. The issue was therefore not LAN isolation or the macOS firewall.

### Root cause

The Rust HTTP client enables reqwest `TlsInfo` and then reads the peer certificate from the response extension in `verify_cert_from_res`. Under the macOS packet-tunnel configuration, the HTTPS request and TLS handshake complete, but that optional response extension is absent. LocalSend consequently rejects an otherwise successful, already-verified TLS exchange.

This response-level check is also redundant for pinned transfer clients because the custom rustls `PinnedServerCertVerifier` already validates the certificate and expected SHA-256 fingerprint during the TLS handshake, before request data is sent.

### Proposed fix and verification

Capture certificates that pass the custom rustls verifier, keyed by TLS server name, and use that verified handshake certificate for discovery identity and legacy public-key checks. This removes the dependency on reqwest response `TlsInfo` without weakening fingerprint pinning.

The change was verified with:

- 7/7 focused TLS-pinning integration tests passing;
- `cargo clippy --features full --all-targets` passing (with existing repository warnings);
- a real encrypted Mac-to-iPhone transfer while the macOS `utun` packet tunnel remained active: discovery, certificate fingerprint pinning, acceptance, and upload all completed successfully.

### Why LocalSend and Shadowrocket are commonly used together

In mainland China, proxy/VPN tools such as Shadowrocket are often kept running continuously because they are needed for normal access to many online services. Therefore, using LocalSend while Shadowrocket is active is not an unusual edge case for Chinese users, but a common everyday setup. LocalSend also has a sizeable Chinese-speaking community, as reflected by the dedicated Chinese-language section in its official Discord. Requiring users to disable their proxy whenever they send a local file would disrupt normal internet access, so LocalSend should remain compatible with macOS packet-tunnel applications while preserving HTTPS certificate verification.

kax168:
## Summary

- capture peer certificates after they pass the custom rustls verifier, keyed by TLS server name;
- use the verified handshake certificate for discovery identity and legacy public-key checks;
- remove the HTTP client's dependency on reqwest response `TlsInfo`, which can be absent behind a macOS packet tunnel.

Fixes #3299.

## Security properties preserved

This does not skip or weaken certificate verification:

- certificate validity is checked before a certificate is recorded;
- pinned transfer clients still compare the expected SHA-256 fingerprint during the TLS handshake, before HTTP request data is sent;
- a fingerprint mismatch is rejected before file metadata reaches the peer;
- upload bodies are not read or transmitted when the fingerprint check fails;
- unpinned discovery derives the public key and fingerprint from the verified handshake certificate rather than trusting identity data in the response body;
- captured certificates remain in memory and are separated by TLS server name.

The response-level certificate check was redundant for pinned clients and unreliable because `reqwest::tls::TlsInfo` is an optional response extension. The custom rustls verifier is the authoritative point at which LocalSend sees and authenticates the peer certificate.

## Testing

- `cargo test --features full --test v2_tls_pinning` (7 passed)
- `cargo clippy --features full --all-targets` (passes with existing repository warnings)
- focused discovery tests passed, including reading the HTTPS identity from the certificate
- real encrypted macOS-to-iPhone transfer with the macOS default route on an active `utun` packet tunnel: registration, certificate fingerprint pinning, receiver acceptance, and upload all succeeded

The same real-world flow failed in LocalSend 1.18.0 with `RsHttpClientError.other(field0: TLS info not found)` when Shadowrocket was active on macOS.

## User context

In mainland China, proxy/VPN tools such as Shadowrocket are commonly kept active for normal access to online services. Using LocalSend and a packet-tunnel application at the same time is therefore an everyday configuration for many Chinese users, not just an unusual edge case. Disabling the tunnel for every local transfer would also interrupt unrelated internet access.

Development note: this bug fix was prepared with AI assistance and validated through source review, automated security regression tests, and a real-device reproduction before and after the change.

### DIFF
--- a/packages/core/src/http/client/mod.rs
+++ b/packages/core/src/http/client/mod.rs
@@ -195,13 +195,20 @@ pub(super) fn upload_body(
 /// custom certificate verifier. reqwest passes such a config straight through,
 /// which means the client certificate and ALPN have to be set here as well:
 /// `identity()` and the HTTP version preference of the builder no longer apply.
+pub(super) struct ConfiguredClient {
+    client: reqwest::Client,
+    peer_certificates: server_cert_verifier::PeerCertificateStore,
+}
+
 pub(super) fn create_reqwest_client(
     private_key: &str,
     cert: &str,
     expected_fingerprint: Option<String>,
     timeout: Option<std::time::Duration>,
-) -> Result<reqwest::Client, ClientError> {
+) -> Result<ConfiguredClient, ClientError> {
     let _ = rustls::crypto::ring::default_provider().install_default();
+
+    let peer_certificates = server_cert_verifier::PeerCertificateStore::default();
 
     let mut tls_config = {
         let certs =
@@ -215,6 +222,7 @@ pub(super) fn create_reqwest_client(
                 server_cert_verifier::PinnedServerCertVerifier::try_new(
                     cert,
                     expected_fingerprint,
+                    peer_certificates.clone(),
                 )?,
             ))
             .with_client_auth_cert(certs, key)
@@ -235,7 +243,10 @@ pub(super) fn create_reqwest_client(
 
     let client = builder.build()?;
 
-    Ok(client)
+    Ok(ConfiguredClient {
+        client,
+        peer_certificates,
+    })
 }
 
 /// DNS resolver that turns the synthetic host names produced by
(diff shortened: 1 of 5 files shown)


Tienisto:
Thank you. A different fix was made on main.

Issue #3060, opened 2026-05-06
bug: Incorrect date (1970) and wrong orientation for imported media on Android 8


EDCrasher:
### Read [this](https://github.com/localsend/localsend/issues/528) before anything

- [ ] I have read it.
- [ ] Issues not fitting are closed.

### Make sure about:

- [x] This issue is not duplicated with any other open or closed issues.
- [x] I am using the latest version from the release.

### Describe the bug

When transferring photos and videos to my Android device, the files are successfully imported and appear in the gallery. However, immediately after transfer, two issues occur:

1. The media files display an incorrect creation date, often showing 1970.
2. Some images appear rotated incorrectly (e.g., sideways or upside down), even though they are correctly oriented on the source device.

After restarting the device, both the correct timestamps and proper image orientation are restored automatically after a short period of time.

### Expected behavior

The date should be displayed correctly, and the image orientation should not be distorted.

### Reproduction

Send a photo or video to a Huawei Mate 10 Lite device.

### Device information

**Device:** Huawei Mate 10 Lite (RNE-L21)
**Operating System:** Android 8.0.0 - EMUI 8.0.0

### Screenshots

<img width="1080" height="801" alt="Image" src="https://github.com/user-attachments/assets/0b672a78-2370-4457-bcc7-fb4d3709296a" />

Issue #2940, opened 2026-02-15
bug: Doesn't keep original date of file


bibsp:
### Read [this](https://github.com/localsend/localsend/issues/528) before anything

- [x] I have read it.
- [x] Issues not fitting are closed.

### Make sure about:

- [x] This issue is not duplicated with any other open or closed issues.
- [x] I am using the latest version from the release.

### Describe the bug

When Sending file from phone to PC the `Date modified` and `Date created` get changed depending on how I picked the file.
1. If file is picked from phone's file manager/gallery app and send it through LocalSend then both the dates get changed.
2. If file is picked from LocalSend's in-built file picker and send it through LocalSend then only the `Date created` gets changed.

### Expected behavior

All the dates should remain unchanged like original file.

### Reproduction

Send Any file from Android to Windows PC.

### Device information

**From:**
Redmi Note 7S
Android 10

**To:**
Windows 10 22H2

### Screenshots

<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/c4716447-370b-4588-a411-e611fe27a0ac" />
<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/09845be9-8051-4d03-a9a4-e15d81dff528" />

Niloyyy:
is anyone working with this? if not then let me know.

bibsp:
@Niloyyy Thanks for your interest.

One note: I only checked one direction trasfer. From Phone to PC.

I don't know if this issue exist in opposite direction as well or not. From PC to Phone.

So if anyone fixes this issue please check in both directions.
Thank you.

bibsp:
Hello guys, any luck on this? This is not a huge feature request. It's a small bug. Can anyone give priority to fix it?

I really want to switch to this app. Only this issue is preventing me to use it.

_(And one more thing. The app is not updated since Feb 2025. What happened to this project? Even if somene fixes this, how will we get latest version? Please release updates frequently.)_

OdyAsh:
Up! For an awesome application like this, this is indeed an irritating issue :(

OdyAsh:
For context, a workaround for this is to either:
1. ZIP the files you want to transfer first (using apps like Solid File Explorer), then sending that ZIP file using localsend, then extracting that ZIP file in the recipient device.
[...]

PR #3334, opened 2026-08-18
fix: do not send a 1970 timestamp when SAF does not know the file's mtime

Shlomo116:
## Problem

Both Android SAF paths read the modification time out of a cursor with `getLong`:

```kotlin
// FastDocumentFile.kt — listFiles() and fromDocumentUri()
lastModified = cursor.getLong(4)
```

`DocumentsContract.Document.COLUMN_LAST_MODIFIED` is [documented as nullable](https://developer.android.com/reference/android/provider/DocumentsContract.Document#COLUMN_LAST_MODIFIED) ("value may be `null` if unknown"), and `Cursor.getLong` returns `0` for a NULL column. Several document providers leave it empty.

Dart then converts that `0` unconditionally:

```dart
lastModified: DateTime.fromMillisecondsSinceEpoch(file.lastModified, isUtc: true).toIso8601String(),
```

So the file goes out over the wire with `lastModified: "1970-01-01T00:00:00.000Z"`. On the receiving side that parses fine and `save.rs` applies it, so the saved file gets an mtime of 1970 — worse than the no-metadata case, where the receiver would simply leave the file alone.

The same conversion is duplicated in `selected_sending_files_provider.dart` for the directory-picker path, so both the file picker and the folder picker are affected.

## Change

Extracted the conversion into `safTimestampToRfc3339` and made it return `null` for a non-positive value, which is already how the protocol expresses "no metadata" — `FileMetadata` skips absent fields and the receiver leaves the timestamp untouched. Both call sites now go through it.

## Verification

- Three unit tests in `app/test/unit/util/native/cross_file_converters_test.dart`, covering the conversion itself, the unknown case, and `convertFileInfo` end to end. The first one also pins the existing RFC 3339 output so the refactor is provably behaviour-preserving for real timestamps.
- `flutter test` — all 72 tests pass.
- `flutter analyze` — no issues.
- `dart format --set-exit-if-changed` — clean.

I do not have an Android device with a provider that omits the column, so the fix is argued from the `DocumentsContract` contract and covered by unit tests rather than reproduced on hardware. Ran against Flutter 3.44.4 rather than the pinned 3.41.9, since I don't have fvm set up here; `app/pubspec.yaml` allows `^3.41.0`.

I did not find an issue that pins this down exactly, though it is a plausible contributor to the scattered "transferred file shows 1970 / wrong date modified" reports (#3060, #2940).

---

This fix was written with Claude Code. Per `CONTRIBUTING.md`, AI-assisted contributions are allowed for bug fixes; flagging it explicitly so you can weigh it as you see fit.

### DIFF
--- a/app/lib/provider/selection/selected_sending_files_provider.dart
+++ b/app/lib/provider/selection/selected_sending_files_provider.dart
@@ -247,7 +247,7 @@ class AddAndroidDirectoryAction extends AsyncReduxAction<SelectedSendingFilesNot
         path: file.uri,
         bytes: null,
         // SAF only provides milliseconds, so there is no point statting in Rust.
-        lastModified: DateTime.fromMillisecondsSinceEpoch(file.lastModified, isUtc: true).toIso8601String(),
+        lastModified: safTimestampToRfc3339(file.lastModified),
         lastAccessed: null,
       );
 
--- a/app/lib/util/native/cross_file_converters.dart
+++ b/app/lib/util/native/cross_file_converters.dart
@@ -69,7 +69,7 @@ class CrossFileConverters {
       path: file.uri,
       bytes: null,
       // SAF only provides milliseconds, so there is no point statting in Rust.
-      lastModified: DateTime.fromMillisecondsSinceEpoch(file.lastModified, isUtc: true).toIso8601String(),
+      lastModified: safTimestampToRfc3339(file.lastModified),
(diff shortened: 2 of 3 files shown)


Tienisto:
Thanks. I moved the fix to Kotlin on main.

bibsp:
@Tienisto is 2940 issue fixed?

PR #3340, opened 2026-08-19
fix: handle mixed files and directories in drag-and-drop

MetehanKacar:
### Problem
When dragging and dropping a mix of files and directories onto the home page at the same time, the app did not accept them correctly. The old code only handled the directory case when a **single** item was dropped and that item was a directory; otherwise it treated **all** dropped items as files. This meant:

- 1 directory + 1 or more files dropped together → the directory was processed as a file and failed
- Multiple directories dropped together → all were treated as files
- Mixed drops were never expanded correctly

### Fix
Rewrote the `onDragDone` handler to split dropped items into **files** and **directories**, then process both groups correctly:

1. For each dropped `XFile`, check `Directory(path).existsSync()`
2. Directories are added via `AddDirectoryAction` (expanded recursively)
3. Files are added directly via `AddFilesAction`
4. Finally, the user is switched to the `Send` tab

### DIFF
--- a/app/lib/pages/home_page.dart
+++ b/app/lib/pages/home_page.dart
@@ -83,20 +83,34 @@ class _HomePageState extends State<HomePage> with Refena {
         });
       },
       onDragDone: (event) async {
-        if (event.files.length == 1 && Directory(event.files.first.path).existsSync()) {
-          // user dropped a directory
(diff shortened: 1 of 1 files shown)


Tienisto:
Thank you. Fixed this issue on main

PR #3355, opened 2026-08-22
fix(linux): set StartupWMClass in the deb desktop entry

vicajilau:
### Problem

The dock and alt-tab show a generic icon for the running window instead of the
LocalSend logo.

### Cause

The Linux app sets its GTK application id to `org.localsend.localsend_app`:

```cmake
set(APPLICATION_ID "org.localsend.localsend_app")   # linux/CMakeLists.txt
```
```cpp
g_set_prgname(APPLICATION_ID);                      # linux/my_application.cc
```

The desktop entry generated for the deb is named `localsend_app.desktop` and
declares no `StartupWMClass`, so nothing maps the window back to it. GNOME
Shell tries, in order, `StartupWMClass` against the WM_CLASS instance and
class, then a desktop file named after them, and finally the GApplication id,
see `get_app_from_window_wmclass` in `shell-window-tracker.c`. Every step looks
for `org.localsend.localsend_app`, which matches nothing.

### Fix

Declaring `StartupWMClass` covers both session types. On Wayland the app id is
`org.localsend.localsend_app`, and on X11 GTK uses it verbatim as the WM_CLASS
instance part, which is the first thing GNOME Shell checks.

The rpm config has no equivalent knob, `flutter_app_packager` only emits
`StartupWMClass` for deb, so the rpm keeps the current behaviour.

### Testing

`dart format` clean, `flutter analyze` reports no issues, `flutter test` passes.

Built the package with `flutter_distributor package --platform linux --targets
deb` and read the entry back out of it with `dpkg-deb -x`:

```desktop
[Desktop Entry]
Type=Application
Name=LocalSend
GenericName=An open source cross-platform alternative to AirDrop
Icon=localsend_app
Exec=localsend_app %U
Categories=GTK;FileTransfer;Utility;
Keywords=Sharing;LAN;Files;
StartupNotify=true
StartupWMClass=org.localsend.localsend_app
```

The value matches the application id compiled into the shipped binary. The
generic icon was observed on Ubuntu 26.04, GNOME on Wayland.

### DIFF
--- a/app/linux/packaging/deb/make_config.yaml
+++ b/app/linux/packaging/deb/make_config.yaml
@@ -39,3 +39,7 @@ categories:
   - Utility
 
 startup_notify: true
+
+# Must match the GTK application id set in linux/CMakeLists.txt, otherwise the
+# desktop environment cannot map the running window back to this desktop entry.
+startup_wm_class: org.localsend.localsend_app


ArqamQazi:
Thanks for PR, I am having the same issue right now. 

Better way to solve this problem would be to change the .desktop file name of localsend to match the appid.

see: https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:set_app_id
> The compositor shell will try to group application surfaces together by their app ID. As a best practice, it is suggested to select app ID's that match the basename of the application's .desktop file. For example, "org.freedesktop.FooViewer" where the .desktop file is "org.freedesktop.FooViewer.desktop".

vicajilau:
Thanks, that is the canonical best practice and I agree it is the better end state. The catch is that in the deb pipeline the entry name is not free.

`flutter_app_packager` derives it from the binary name rather than the package name:

```dart
final desktopEntryFile =
    File(path.join(applicationsDir, '${makeConfig.appBinaryName}.desktop'));
```

and the same `appBinaryName` drives the rest of the package:

```dart
'ln -s /opt/$appBinaryName/$appBinaryName /usr/bin/$appBinaryName',
'Icon': appBinaryName,
'Exec': '$appBinaryName %U',
```

So renaming the entry to `org.localsend.localsend_app.desktop` means renaming the executable too, which moves the install prefix to `/opt/org.localsend.localsend_app/`, changes the command on `$PATH` and changes the icon name. That is a packaging identity change rather than an icon fix, and it breaks launchers pinned to the current entry. The Flatpak already ships under the app id, which is exactly why it does not have this bug, so aligning the others is a reasonable goal.

For what it is worth, the AppImage packaging hit the same bug and landed the same fix a day ago: pkgforge-dev/LocalSend-AppImage@c418fc4 sets `STARTUPWMCLASS=org.localsend.localsend_app`.

**So, concretely: I suggest merging this as is and treating the rename as a separate decision for the maintainers.** This fixes the icon for deb users now, in one line, with no change to the install layout or the launch command. The two are not mutually exclusive either, so if the entry is renamed to match the app id later, this line just becomes redundant and can be dropped.

AcideFluorhydrique:
Confirming this fix works.

**Environment:** Ubuntu 26.04.1 LTS, GNOME Shell 50.1, Wayland, LocalSend installed from the official `.deb`

**Regression:** The dock showed a generic gear icon for the running window starting with 1.18.0 (1.17.0 was fine), and it's still broken in 1.18.2. This matches `g_set_prgname(APPLICATION_ID)` being added in d67b84f, which first shipped in 1.18.0. That changed the app id from `localsend_app` to `org.localsend.localsend_app`, and the desktop file was never updated to match.

**Workaround / verification:** I copied `localsend_app.desktop` into `~/.local/share/applications/` and added:

    StartupWMClass=org.localsend.localsend_app

After restarting LocalSend, the dock shows the correct icon, which is exactly what this PR does.

Related: #3416

PR #3360, opened 2026-08-22
Fix Android system name resolving to localhost

### LINKED ISSUE #3357: bug: 在LocalSend Android 1.18.2版本应用中，尝试点击显示器图标快捷更改设备名称会更改为localhost，而非当前手机型号

- [x] I have read it.

### Make sure about:

### Describe the bug

在LocalSend里面的设置→网络→设备名称。里面有一个快捷设置名称为当前设备名称/型号（图标是显示器）的。但是我为什么在Android App上点击这个图标，设备名称会被更改为localhost

### Expected behavior

通过这个方式更改设备名称，应该更改为当前的设备型号，或者是在手机设置里手动设置的设备名称

### Reproduction

1.打开LocalSend
2.在下方按钮中找到设置
3.往下找，找到网络设置
4.找到设备名称，点击名称试图更改它
5.点击窗口标题下面的显示器图标

### Device information

设备：HUAWEI Mate 30 Pro 5G
系统：Android12

### Screenshots

_No response_

### Additional context

_No response_

WhiteJadeWolf:
## Summary

Fixes #3357.

On Android, `Platform.localHostname` can resolve to `localhost` instead of the device's actual name.

This change uses Android's device information to generate the system name from the device brand and model, while avoiding duplicate brand names when the model already includes it.

## Testing

Tested the "Use System Name" action on:

- A Nothing Phone (2)
- An Android emulator

The "Use System Name" action now uses the Android device's brand and model instead of resolving to `localhost`.

### DIFF
--- a/app/lib/pages/tabs/settings_tab.dart
+++ b/app/lib/pages/tabs/settings_tab.dart
@@ -1,5 +1,6 @@
 import 'dart:io';
 
+import 'package:device_info_plus/device_info_plus.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:localsend_app/config/theme.dart';
@@ -386,6 +387,11 @@ class SettingsTab extends StatelessWidget {
                             if (Platform.isMacOS) {
                               final result = await Process.run('scutil', ['--get', 'ComputerName']);
                               newAlias = result.stdout.toString().trim();
+                            } else if (Platform.isAndroid) {
+                              final deviceInfo = await DeviceInfoPlugin().androidInfo;
+                              final brand = deviceInfo.brand;
+                              final model = deviceInfo.model;
+                              newAlias = model.toLowerCase().startsWith(brand.toLowerCase()) ? model : '$brand $model';
                             } else {
                               newAlias = Platform.localHostname;
                             }


Richard19Perez77:
Nice fix — this looks like the right approach for the Android localhost case, and the brand/model dedupe is a nice touch.

Two small optional ideas:

1. Tests are expected for LocalSend code changes. Pulling the name logic into a tiny helper would make that easy, something like:

```dart
String androidSystemName({required String brand, required String model}) {
  return model.toLowerCase().startsWith(brand.toLowerCase())
      ? model
      : '$brand $model';
}
```

Then the button becomes roughly:

```dart
} else if (Platform.isAndroid) {
  final info = await DeviceInfoPlugin().androidInfo;
  newAlias = androidSystemName(brand: info.brand, model: info.model);
}
```

2. iOS still falls through to `Platform.localHostname`. If that can also be `localhost`, maybe handle it later (or keep this PR Android-only on purpose).

Core Android fix looks good either way.

WhiteJadeWolf:
Thanks for your input.

> Pulling the name logic into a tiny helper would > make that easy, something like:
> 
> ```dart
> String androidSystemName({required String brand, required String model}) {
>   return model.toLowerCase().startsWith(brand.toLowerCase())
>       ? model
>       : '$brand $model';
> }
> ```
> 
> Then the button becomes roughly:
> 
> ```dart
> } else if (Platform.isAndroid) {
>   final info = await DeviceInfoPlugin().androidInfo;
>   newAlias = androidSystemName(brand: info.brand, model: info.model);
> }
> ```

I didn't feel the need to write a helper for something which can be done on the fly, especially when the supposed helper function will only be used once (in this case).


> 2. iOS still falls through to `Platform.localHostname`. If that can also be `localhost`, maybe handle it later (or keep this PR Android-only on purpose).

The issue explicitly mentioned that its a bug corresponding to Android, so i submitted a PR which solves exactly that and did not mess with the rest of the code which might lead to unnecessary edge cases.

"if its not broken, don't try to fix it" :)

PR #3370, opened 2026-08-25
fix(web): don't let empty files kill the whole browser upload

### LINKED ISSUE #3140: bug: 0KB zip file is wrongly included in the transmission and stops sending; the zip filename starts with some of the IMG files and is like "IMG_XXXX.pvc.zip"

- [x] I have read it.
- [x] Issues not fitting are closed.

### Make sure about:

### Describe the bug

When sending many files from iOS Files app to PC, on the receiving side, I see some unexpected "IMG_XXXX.pvc.zip" of 0 KB of size, and the process stops there, I have to start all over. 

Maybe it's due to the IMG_XXXX.pvc.zip shares the same starting part of the filename with one of the IMG file I want to share; maybe the empty file is created by some failed effort of zipping all the files I'd like to share, due to no space left on disk. But LocalSend should not select the zip file in any case, when I selected some real image files from Files. The zip file is never visible to me in Files app to be selected.

### Expected behavior

When I select multiple image files from Files app in iOS 17, without selecting any zip file in the same dir(actually the zip file is not even visible for me), I expect LocalSend to select the real image files, instead of the 0KB zip file named "IMG_XXXX.pvc.zip", no matter if the starting part of the filename is the same as one of the image I want to send; and no matter if there's any zip file, as I was selecting image files to send. If such zip files are there, they should be skipped and ignored, and/or the whole process should not stop due to that error. 

### Reproduction

1. Fill your storage to almost full on iOS
2. Put many image files in your Files app, under some dir
3. Compress many image files in that dir, but immediately stops(or let it stop, because there's no enough space on the disk)
4. You will not see the zip file in Files app, as it failed. But I guess the zip file is named after the first image I selected to compress, thus IMG_XXXX blah blah blah
5. Start using the browser app in Chrome
6. Select the same image files you tried to compress to send to your PC, also Chrome browser version
7. Some files will be received, but in the receiving end(PC), you will find one "IMG_XXXX.pvc.zip" file of 0KB, and once reaching it, the process will stop, so you only received part of the files and have to start again.

### Device information

iphone 12, iOS 17, Chrome, webapp
to a PC, Win 11, Chrome, webapp

### Screenshots

_No response_

### Additional context

On the sending end, the image IMG_XXXX(with same the name as the erroronous zip filename first part) now is not png, but a zip file, but it can still be viewed in Files as image, and I can open it as image. I can export it to Photos and save it back to Files, then it’s an image again, then it can be sent normally.

loucass:
### What
Browser sends via the receive-via-link page aborted the entire queue when any
single file failed — e.g. the 0 KB corrupted zip from #3140 that iOS silently
includes with selections sharing its filename prefix.

- filter zero-byte files before prepare-upload
- per-file failures (500 / connection lost) are counted and keep uploading;
  the end status shows "N/M file(s) failed"
- session-level failures (403/409/404) still abort immediately, since retrying
  cannot succeed once the session is gone
  
  Fixes #3140

### DIFF
--- a/packages/core/assets/web/upload.html
+++ b/packages/core/assets/web/upload.html
@@ -201,11 +201,16 @@
       hideProgress();
       showStatus(i18n.waiting);
 
+      // Zero-byte files (e.g. corrupted leftovers that the browser includes
+      // with a selection, #3140) would abort the whole transfer midway.
       var selectedFiles = {};
       var filesDto = {};
       for (var i = 0; i < fileList.length; i++) {
         var file = fileList[i];
-        var fileId = String(i);
+        if (!file.size) {
+          continue;
+        }
+        var fileId = String(Object.keys(selectedFiles).length);
         selectedFiles[fileId] = file;
         filesDto[fileId] = {
           id: fileId,
@@ -213,6 +218,11 @@
           size: file.size,
           fileType: file.type || 'application/octet-stream'
         };
+      }
+
+      if (Object.keys(selectedFiles).length === 0) {
+        finishWithError(i18n.nothingToSend || 'Nothing to send');
+        return;
       }
 
       var body = JSON.stringify({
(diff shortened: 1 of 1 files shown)


loucass:
Friendly ping @Tienisto three PRs ready when you have a moment: #3370, #3371, #3373. Happy to rebase or split as needed, no rush.

PR #3374, opened 2026-08-27
fix: show "LocalSend" as title in the Linux system tray

Josephur:
## Problem

On Linux desktops (I ran into this on KDE Plasma 6 / Wayland), the system tray "Status and Notifications" popup shows the raw application ID `org.localsend.localsend_app` instead of "LocalSend".

## Root cause

The tray label is the StatusNotifierItem `Title` property. `libayatana-appindicator` fills `Title` from `g_get_application_name()`. Flutter's Linux runner never calls `g_set_application_name()`, so GTK falls back to `g_get_prgname()` — which `my_application_new()` sets to `APPLICATION_ID`. The app ID therefore ends up as the tray label.

## Fix

Set the human-readable application name next to the existing `g_set_prgname()` call in `my_application_new()`:

```c
g_set_application_name("LocalSend");
```

## Before / after

Screenshot_20260825_224408
Screenshot_20260826_060251

### DIFF
--- a/app/linux/my_application.cc
+++ b/app/linux/my_application.cc
@@ -112,6 +112,10 @@ MyApplication* my_application_new() {
   // corresponding .desktop file. This ensures better integration by allowing
   // the application to be recognized beyond its binary name.
   g_set_prgname(APPLICATION_ID);
+  // The application name is shown instead of the program name wherever a
+  // human-readable label is expected, e.g. as the title of the system tray
+  // entry on Linux desktops.
+  g_set_application_name("LocalSend");
 
   return MY_APPLICATION(g_object_new(my_application_get_type(),
                                      "application-id", APPLICATION_ID,


Josephur:
Adding a note I should have included up front: this patch was written with AI assistance, and this comment is being posted through it as well. I'm aware of the AI policy in `CONTRIBUTING.md` and I think this lands inside both exemptions — it's a bug fix, and it's about as small as a change gets: one `g_set_application_name()` call plus a three-line comment, in one file, with no behavioural surface beyond the tray label. The app ID, program name and D-Bus single-instance registration all still use `APPLICATION_ID` and are untouched.

More to the point, I haven't just generated it and thrown it over the wall — I've been running this patch as my daily LocalSend build for a few weeks now with no ill effects. Tray label reads "LocalSend", and single-instance handover, the window title and the `.desktop` association all behave exactly as before.

Happy to drop the comment or reword the commit if you'd like it leaner.

Issue #2105, opened 2024-11-28
Image preview generation is very slow


Nerdoholic:
### Describe the bug

When selecting multiple images to send, you generate a preview for each.

This is a nice feature, but
1) I don't care for that feature, my PC is just a transfer dump for my phone (which is already organized, pre-selected via the file menu or simply a "dump all, sort later" for DCIM)
2) it is very slow. Pixel 8a, generating 10 images takes several seconds.

So I would love the simple ability to turn these off.
I have a preview in my gallery app for picking via "Files", that works great.

For others who might work with the preview feature: It shouldn't be that slow.
Do not though: This might be a Pixel 8a issue of some sort. I reported to Google that their USB preview (Windows viewing DCIM) is the slowest POS in the universe (1 image per second, I kid you not) and they couldn't provide a fix! So maybe it's hardware dependent.
Just adding this 2nd "bug flow direction" it in case others do use the feature and are affected by it.

(1) is a good thing to have in general - I'd be happy if this could be implemented.

If I can help by pulling some logs or whatever, let me know.

PS: The app is amazing!
KDE Connect recently had big issues and it was a blessing to find this!!

### To reproduce

Select images to be transfered via the "File" menu. The preview usually takes 5-10 seconds, which is an annoying blocker in one's workflow.

### Expected behavior

Image preview (esp. lowres) is <1s aka "instant".

### Desktop (please complete the following information)

W11 x64 WiFi 6E
LocalSend 1.16.1

### Smartphone (please complete the following information)

Pixel 8a
Andoid 15
Build AP3A.241105.007

TomCrush:
I'm here to ask for the same thing: The option to disable thumbnail generation for files being sent.

In my case, I'm copying a directory of almost 1 TB, and I don't need to see the thumbnails, because I'll just leave the program working and I already know what files I'm sending.

Minimized, I have LocalSend using almost 50% of CPU, just because it's apparently generating the thumbnails.

Otherwise, it's a great tool.

LorenVS:
I spent a bit of time digging into this. I'm on a linux desktop, but was able to force thumbnails to go through the `_ContentUriImage` code path by translating the file paths into file URIs. Once I did that, I was able to reproduces a definite slow-down of the preview rendering.

[...]

PR #3395, opened 2026-09-02
perf(android): avoid copying files when selecting, sharing and previewing

V2arK:
## Problem

On Android, three separate paths read or copy a file in full before it can be sent, which makes adding a handful of large photos or videos take tens of seconds or even minutes (sometimes crash before finishing). Reproduces reliably with a dozen large videos.

1. **Media picker.** `CrossFileConverters.convertAssetEntity` uses `AssetEntity.originFile`, which makes photo_manager copy every asset byte-for-byte into the app cache (`ScopedCache.getCacheFileFromEntity` on Android 10+). Selecting a few GB of video means copying a few GB before anything is queued.
2. **Share intents.** `share_handler` copies every shared URI into the app cache too (`FileDirectory.getAbsolutePath`), and does it synchronously from `onNewIntent`, so the copy blocks the main thread. Its `Dispatchers`/`withContext` imports are unused.
3. **Thumbnails.** `_ContentUriImage` reads the whole file into memory just to render a 64px preview (#2105).

None of these copies are necessary: uploading, hashing and web downloads already resolve `content://` through a file descriptor (`FileContent::Fd`), so a content URI is all the app needs to hold on to.

## Changes

One commit each, independent:

- **Media picker** — use the MediaStore content URI as the path instead of `originFile`. `MediaStore.setRequireOriginal()` preserves the location EXIF tags and is gated on `ACCESS_MEDIA_LOCATION`, because opening an original URI without that permission throws. Redaction zero-fills in place, so the declared size stays correct whether or not the permission is granted. photo_manager reports a size of `0` when it cannot reach an asset, so that case falls back to the copy rather than offering the peer an empty file.
- **Share intents** — `MainActivity` claims file-carrying shares and answers with content URIs; text-only shares still go through share_handler. The launch intent has to be claimed before the plugins attach, since share_handler reads and copies it during `onAttachedToActivity`. Only the openable columns are queried, as those are the ones every content provider must expose for a shared URI. A share is claimed all or nothing: handing over just the readable subset would drop the rest without the user ever learning about it.
- **Thumbnails** — ask `ContentResolver.loadThumbnail()` off the main thread, falling back to the previous full read below Android 10 (`minSdk` is 24). Videos get a preview for the first time on this path, and never fall back to the full read, which for a video would cost gigabytes and still not decode as an image.

`lastModified` is now `null` for shared files: share intents do not carry one, and the openable columns do not expose one. Previously this field held the time the cache copy was made, not the file's own timestamp.

Everything is gated on Android (`checkPlatform`) or on the `content://` scheme, so iOS and desktop behaviour is unchanged. iOS still copies through `originFile`; that needs a different fix and is not part of this PR.

## Test plan

Automated, on Flutter 3.41.9 as pinned in `.fvmrc`:

| Command | Result |
| --- | --- |
| `fvm flutter analyze` | No issues found |
| `fvm dart format --set-exit-if-changed lib test` | 286 files, 0 changed |
| `fvm flutter test` | 71 passed |
| `fvm flutter build apk --debug --target-platform android-arm64` | Built |
| `./gradlew :app:lintDebug` | Two errors, both pre-existing in `AndroidManifest.xml` (`ProtectedPermissions`, `QueryAllPackagesPermission`); nothing new |

Analyze and the APK build were run at each of the three commits individually, so each one compiles on its own.

Manual testing environment:

| | |
| --- | --- |
| Device | Samsung SM-F966W (Galaxy Z Fold 7) |
| OS | Android 16, SDK 36 |
| Build | `1.18.2+64`, debug, `android-arm64` |
| Relevant plugins | photo_manager 3.12.0, share_handler_android 0.0.11, wechat_assets_picker 10.1.3, uri_content 3.1.1 |
| Permissions | `READ_MEDIA_IMAGES`, `READ_MEDIA_VIDEO` and `ACCESS_MEDIA_LOCATION` all granted |

| Case | Result |
| --- | --- |
| Media picker, a dozen large videos | Returns instantly; previously tens of seconds |
| Share from gallery, a dozen large videos | No lag; previously an apparent freeze |
| Share a single image (`ACTION_SEND`) | Works |
| Share several images (`ACTION_SEND_MULTIPLE`) | Works |
| Share while the app is running | Works |
| Share while the app is not running | Works |
| Thumbnails, images and videos | Appear immediately; videos previously had none |
| Location EXIF through the media picker | Preserved; latitude, longitude and altitude all intact on the received file |
| Byte integrity | A 4.6 MB Samsung motion photo arrived complete: EXIF, the embedded MP4 (`ftyp`/`moov`/`mdat`) and the trailing vendor `SEF` block all survived |

Not covered, and worth a reviewer's attention:

- The path taken when `ACCESS_MEDIA_LOCATION` is denied. Android grants it silently alongside the media permission rather than prompting separately, and `_pickMedia` re-requests it on every pick, so revoking it through `adb` is undone as soon as the picker opens. Only the granted case was measured.
- Text-only sharing still routes through share_handler by design, but was not exercised.
- The pre-Android 10 fallbacks (no `loadThumbnail`, no `setRequireOriginal`) were not run; the test device is Android 16.

## Note

Shared content URIs carry a temporary grant rather than a persistable one, so a selection made from a share is only valid while the activity lives. That matches the current behaviour: the engine, isolates and in-memory selection die with the activity anyway, and the cache is cleared on every app start.

### DIFF
--- a/app/android/app/src/main/kotlin/org/localsend/localsend_app/MainActivity.kt
+++ b/app/android/app/src/main/kotlin/org/localsend/localsend_app/MainActivity.kt
@@ -8,26 +8,40 @@ import android.content.Context
 import android.content.Intent
 import android.content.pm.PackageManager
 import android.database.Cursor
+import android.graphics.Bitmap
 import android.net.Uri
 import android.os.Build
 import android.os.Environment
+import android.os.Handler
+import android.os.Looper
 import android.provider.DocumentsContract
 import android.provider.MediaStore
 import android.provider.OpenableColumns
 import android.provider.Settings
 import android.util.Log
+import android.util.Size
+import androidx.annotation.RequiresApi
 import io.flutter.embedding.android.FlutterActivity
 import io.flutter.embedding.engine.FlutterEngine
 import io.flutter.plugin.common.MethodCall
 import io.flutter.plugin.common.MethodChannel
+import java.io.ByteArrayOutputStream
 import java.text.SimpleDateFormat
 import java.util.Date
 import java.util.Locale
 import java.util.TimeZone
+import java.util.concurrent.Executors
 
 
 private const val TAG = "LocalSend"
 private const val CHANNEL = "org.localsend.localsend_app/localsend"
+
+// The thumbnails are only ever shown in a small box, so the artefacts stay invisible.
+private const val THUMBNAIL_QUALITY = 80
+
+// Decoding a selection is IO bound, so the pool only has to keep the provider busy.
+private val thumbnailThreads = Runtime.getRuntime().availableProcessors().coerceIn(2, 4)
+
 private const val REQUEST_CODE_PICK_DIRECTORY = 1
 private const val REQUEST_CODE_PICK_DIRECTORY_PATH = 2
 private const val REQUEST_CODE_PICK_FILE = 3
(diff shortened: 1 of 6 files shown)


V2arK:
@Tienisto This is bothering me for a very long time, and I veified on my phone it's working

Issue #3146, opened 2026-06-23
Share via LocalSend Option not proceeding further actions.


atharvyadav22:
### Read [this](https://github.com/localsend/localsend/issues/528) before anything

- [x] I have read it.
- [ ] Issues not fitting are closed.

### Make sure about:

- [x] This issue is not duplicated with any other open or closed issues.
- [x] I am using the latest version from the release.

### Describe the bug

When clicked Share Via "LocalSend" nothing shows

https://github.com/user-attachments/assets/d278d324-7a0e-4976-b8d9-7fbb02aae665

### Expected behavior

Ideally when clicked Share via "LocalSend" app should proceed with opening and redirect further actions

### Reproduction

1. Right Click any File 
2. Click Share
3. Click LocalSend

### Device information

Device: Mac
System: Macbook Pro M5
Version: Tahoe

Issue #2023, opened 2024-11-08
Feature request


mavidser:
### Is your feature request related to a problem? Please describe.

No. Just a clear Usability upgrade.

### Describe the solution you'd like

Add as a macOS extension so LocalSend can be visible in the 'Share' context menu in finder. This can open the device picker in Localsend.

### Describe alternatives you've considered

The current method of opening the app and sending the file.
I can look into building this feature - but want to check the opinion on maintainers

Tienisto:
Do you use the latest version?

mavidser:
Ah, I updated to latest, and a share target is there. Thanks!

Though, LocalSend doesn't remember the file it launched for, and starts in the default Receive tab, when opened from Share menu.

j-lakeman:
I can confirm that. Thoughts @Tienisto?

Tienisto:
It does remember the shared file on my side.

Maybe related: Can you drop a file into the menu bar icon?

j-lakeman:
You mean drop it into the LocalSend window?

ShlomoCode:
@Tienisto I can confirm that this happens to me with the released app, while the development build works. Dragging to the menu bar icon still works.
After a little investigation, it seems to be related to a shared group not working properly, the `ShareExtention` stores the new file in:
```
/Users/shlomo/Library/Containers/org.localsend.localsendApp.ShareExtension/Data/Library/Preferences/3W7H4PYMCV.localsend.shared_group.plist
```
While the main app accesses:
```
/Users/shlomo/Library/Containers/org.localsend.localsendApp/Data/Library/Preferences/3W7H4PYMCV.localsend.shared_group.plist
```

ShlomoCode:
I see that dragging to the Dock icon doesn't work either, so maybe it's not related to the shared group 🤔 
@Tienisto Does dragging to the Dock icon works for you?

Tienisto:
@ShlomoCode No, dragging a file into the Dock icon somehow doesn't work (anymore?) 💀

70sh1:
May I ask if a fix for this issue is planned? I would really appreciate this feature.

ShlomoCode:
> [@ShlomoCode](https://github.com/ShlomoCode) No, dragging a file into the Dock icon somehow doesn't work (anymore?) 💀

[...]

PR #3411, opened 2026-09-08
fix(macos): restore Finder sharing in DMG builds

vdmkotai:
In the DMG version of LocalSend 1.18.2, choosing a file in Finder and clicking Share → LocalSend opens the app, but the file never appears in the send list. If LocalSend is already running, nothing happens.

The Share Extension passes files to the app through shared preferences. The DMG packaging script breaks this connection when it re-signs the app: it replaces the permissions generated by Xcode with the original template. The shared group name ends up containing the literal `$(AppIdentifierPrefix)` instead of the developer's prefix. The extension then saves the file references in its own private preferences, where the main app cannot read them.

This PR preserves the permissions Xcode generated for each executable and signs the embedded components before the main app. It also checks that the app and extension have access to the same shared group and stops packaging if signing fails. The new script uses Python 3's standard library.

I tested a separately installed copy of 1.18.2 with corrected signing permissions on macOS 26.6.2. Sharing from Finder now adds the file to the send list, both while the app is running and after fully quitting it. The test copy has its own name and bundle identifiers; the machine code matches the original.

Other checks:

- All eight regression tests pass on Python 3.9.6. They cover separate permissions for each component, signing order, invalid group names and signing failures. They also run in the packaging CI job.
- Signature verification passes for the corrected app. The new group check rejects the original release.
- Shell syntax and diff checks pass.

I have not tested a fresh build from source, Apple notarization, or a transfer to another device. Those still need verification before release.

Related to #3146 and the [earlier investigation in #2023](https://github.com/localsend/localsend/issues/2023#issuecomment-2544247906).

Prepared with AI assistance and manually tested.

Screen recording of the local test:

https://github.com/user-attachments/assets/ae63ff8b-e581-406f-b6fa-f22b4af2c5ee


ShlomoCode:
Thanks for investigating this! Closing this PR in favor of #3415, which uses a smaller, direct fix based on your findings. I've credited you as co-author.

